PR #15, opened 2026-01-15
Remove the dependency on private midnight-contracts

kmillikin:
This was a set of example contracts that were used for testing.


kmillikin:
I'm abandoning this change.  Instead we will try to move the `midnight-contracts` into this repository.

PR #24, opened 2026-01-21
Fix `ChargedState` bug

JosephDenman:
This is a bug fix for incorrect initialization of contract state in the JS code.  See #34.


JosephDenman:
Done

kmillikin (approved):
Approved with a suggestion to use the public issue in the changelog (no better time to start than now).

PR #42, opened 2026-01-26
Fix excessive compile times for some loops and some inaccuracies in the witness-disclosure error messages

dybvig:
Addresses Issue 23.  See the change log for details.


compiler/analysis-passes.ss
```diff
@@ -4758,88 +4883,113 @@
...
+      ; whether the first argument is <, >, or = to the second.  list<? could be written
+      ; to use a simple #t/#f less-than predicate, but it would have to call it twice
+      ; for every list element, which would be more expensive for expensive comparisons.
+      (define (list<? elt-compare x1* x2*)
+        (let ([n1 (length x1*)] [n2 (length x2*)])
+          (or (fx< n1 n2)
+              (and (fx= n1 n2)
+                   (let loop ([x1* x1*] [x2* x2*])
+                     (and (not (null? x1*))
+                          (let ([x1 (car x1*)] [x2 (car x2*)])
```
JosephDenman: Looks like `x1` and `x2` are bound but never used.

JosephDenman (changes requested):
I can't find anything wrong with the logic of the program. Just a couple more comments on readability.

PR #66, opened 2026-02-04
Remove the syntax for external circuits

dybvig:
This PR removes the syntax for external circuits.  The compiler now injects natives directly into the standard library module, which is simpler and gives us a single source of truth for natives.

Review notes: The PR eliminates a lot of code for abstracting native argument and return types and checking the circuit declarations against them.  It also changes how the typescript generator detects type-parameterized arguments and return values.  Aside from that, the PR replaces uses of external (circuit) with native (circuit).


compiler/standard-library.compact
pataei: We have https://github.com/LFDT-Minokawa/compact/blob/main/doc/api/CompactStandardLibrary/exports.md where we explain standard library.  Now that native circuits are injected into standard library I'm wondering if we should reorganize the exports.md to have all the native circuits under a new section that explains even though one doesn't see these circuits in the standard library they can still use it since they are internally injected into standard library.

PR #75, opened 2026-02-05
Kmillikin/adjusted release

kmillikin:
I've created this pull request by:

1. reverting the squash commit for the merge from `main` on `release/midnight` (this does not revert cleanly, but almost does)
1. reverting the squash commit for the merge from `release/midnight` on `main`
1. merging `main` into `release/midnight` (this does not merge cleanly, but almost does)

We will still need one more merge back to `main` at the end to ensure that we can merge `main` back into the release branch the next go around (I think).


.github/workflows/release.yml
```diff
@@ -37,18 +37,13 @@ on:
...
-        type: boolean
-        default: true
 
   schedule:
     - cron: '0 */12 * * *'
 
 concurrency:
   group: release-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
```
pataei:
```suggestion
  cancel-in-progress: true
```

kmillikin: Maybe we want: https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-only-cancel-in-progress-jobs-on-specific-branches

kmillikin: One of the reasons that the squash commits did not revert cleanly is because they also contained changes that were not simply part of the merge.  So we should avoid doing that --- I don't want to include this in the merge commit.

It's wrong on main and it should be fixed there.  I'm not going to cancel this build so I don't care that it's wrong for now on the release branch.

Issue #41, opened 2026-01-23
Add flags to the Compact compiler to show the ledger and Compact runtime versions


kmillikin:
The Compact compiler currently targets a single Midnight ledger version and requires a compatible Compact runtime JS package.  You can find the if you know where to look by looking at the source code for the GitHub tag for the release.  It's not possible to get these from the compiler binary itself.

We should add flags to the compiler to print them out.  As a suggestion:

* `--ledger-version` prints the ledger version.  We can get this at build time from `flake.nix`.  Consider that we plan to eventually support multiple ledger versions in the same release, so this might eventually be a list or range of versions.  Also consider how we would spell the flag to select ledger versions.  Is it too confusing if `--ledger=` sets the version and `--ledger-version` prints it?
* `--runtime-version` prints the Compact runtime version that the generated code will import.  I think this is already available to us (we emit it into the generated JS code).

dybvig:
Perhaps the compiler should also extend `pragma` to allow specification of the required `runtime` and `ledger` versions.

dybvig:
Also perhaps the compiler should support a separate flag causing it to print all of the versions it knows in some human-readable format.

kmillikin:
Those are interesting ideas.  I did think that eventually (when the compiler supported multiple ledger versions) that the flag `--ledger-version` would show something like:

```
$ compact compile --ledger-version
Ledger versions: 7.0, 7.1, 8.0
```

Pragmas are interesting.  The contract itself doesn't depend on the version in the same way that it does on the language version (because it needs specific features) or the compiler version (because it needs specific bug fixes).  I don't think it depends on the runtime version at all, and depends only indirectly on the ledger version (because it needs a ledger with certain features, but those features are directly represented in the language version).

So I think we could do that but to be useful the compiler would use it as a way to select from among multiple ledger versions (`pragma ledger_version 7.1` would be a replacement for `--ledger=7.1` on the command line).  Do we want that?

PR #80, opened 2026-02-05
Add a `--runtime-version` flag to the compiler binary

kmillikin:
This is only one half of #41.

### DIFF
--- a/compiler/compactc.ss
+++ b/compiler/compactc.ss
@@ -40,8 +40,9 @@ The following flags, if present, affect the compiler's behavior as follows:
 
   --language-version prints the language version and exits.
 
-  --runtime-version prints the version of the Compact runtime JavaScript package
-    that is used by generated contract code.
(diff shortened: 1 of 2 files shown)


compiler/compactc.ss
```diff
@@ -28,10 +28,10 @@
   (fprintf (current-output-port) "
 This program compiles the Compact source program in the file specified by
 <source-pathname> and puts the resulting target files into the directory
-specified by <target-directory-pathname>.  These target files include
-a Typescript equivalent of the Compact source file, Zkir circuit equivalents
-of the exported circuits, and proving keys created by running zkir on the
-zkir circuits.
+specified by <target-directory-pathname>.  These target files include a
```
kmillikin: It looks like this was wrapped by hand, I couldn't figure out the desired column width.  I used 80 to (have emacs) reformat it.

dybvig (approved):
LGTM with one suggestion regarding the help text for `--runtime-version`.

Issue #112, opened 2026-02-13
Remove one of `NativePoint` or `JubjubPoint` from the standard library


kmillikin:
We don't need two types for this, it complicates things and doesn't provide very much benefit as it is currently implemented.

Remove the `new type` alias (called `NativePoint`) and keep the other `type` alias (currently called `JubjubPoint`).  We can sort out the exact naming of the remaining type before we branch on March 4 for the 0.30 toolchain release.

PR #113, opened 2026-02-13
Remove `NativePoint`, make `JubjubPoint` a `new type` alias

kmillikin:
Closes issue #112.


kmillikin:
> Looks right, but please add a change-log entry and bump the language and compiler version numbers. I don't remember if there was a challenge with fixup renaming types, but I agree it would be nice to have.

Thanks for the reminder.  Done!

PR #121, opened 2026-02-16
Pin some more actions to commit hashes in our workflows

kmillikin:
We are pinning actions to full commit hashes to avoid an "attack" where a tag or branch is made to point elsewhere by a bad actor.

Pin commit hashes for:

- https://github.com/cachix/install-nix-action to v31.9.0
- https://github.com/EnricoMi/publish-unit-test-result-action to v2.22.0
- https://github.com/dtolnay/rust-toolchain to the head of the `stable` branch


kmillikin:
> I still see a "[LOW] Unpinned Actions Full Length Commit SHA" warning over a few of the actions. Should these also be pinned? Or low priority?

I think these are all the same priority, I'm just leaving some for others to fix :)

PR #138, opened 2026-02-17
Change the compiler to target midnight-ledger version 8

kmillikin:
The compiler now targets ledger version 8.0.0, and the Compact runtime now imports on-chain runtime 3.0.0.


kmillikin:
The ledger build issue should be fixed by https://github.com/midnightntwrk/midnight-ledger/pull/203.

PR #154, opened 2026-02-19
Add 'ProvableCircuits' object

JosephDenman:
Fix for issue #150. See issue for problem details.


compiler/typescript-passes.ss
```diff
@@ -1674,7 +1714,20 @@
               '()
               xpelt*))
 
-          (define (print-contract-constructor xpelt0* uname* impure-name*)
+          (define (build-circuit-name-object prefix name*)
+            (apply make-Qconcat
+              (format "this.~a = {" prefix)
+              (if (null? name*)
```
dybvig: I think you could use `Qsep` here.  Like `(apply (Qsep ",") (map --- name*))` where `---` is probably just `(lambda (name) (format "~a: this.circuits.~;*~a" name))`.

dybvig (approved):

Issue #103, opened 2026-02-12
dependabot needs to be able to run our CI


kmillikin:
We have dependabot enabled and we get findings submitted as PRs.

Unfortunatly, dependabot cannot run our CI tests, so these PRs are not fully tested and so we don't have any confidence that we can land them without breaking things we care about.

We need to give dependabot access to our CI or else have a really simple documented process for how to test and land such PRs.

JosephDenman:
This should be fixed in #160 .

PR #160, opened 2026-02-20
Split test reporting into separate workflow for fork PR support

JosephDenman:
Fork PRs run with a read-only `GITHUB_TOKEN`, which causes the test reporting steps (`dorny/test-reporter`, `EnricoMi/publish-unit-test-result-action`, `github-actions-ctrf`) to fail with 403 errors when posting PR comments. This PR moves test reporting into a new `workflow_run`-triggered workflow that runs from the default branch with write permissions, while the build workflow uploads test results as artifacts. This follows the two-workflow pattern [recommended](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) by GitHub Security Lab to safely handle fork PRs without exposing secrets.

### DIFF
--- a/.github/workflows/publish-test-results.yml
+++ b/.github/workflows/publish-test-results.yml
@@ -16,15 +16,16 @@ on:
     types:
       - completed
 
-permissions:
-  checks: write
-  pull-requests: write
-  actions: read
+permissions: {}
 
 jobs:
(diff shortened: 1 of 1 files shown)


kmillikin:
This may fix #103.  If so, we should close the issue and link to the PR.

PR #167, opened 2026-02-25
Add a JubjubScalar type to the standard library

kmillikin:
`JubjubScalar` is new Compact foreign field type.  It has casts to and from `Field` and all `Uint` types:

- casts from `Field` or `Uint` to `JubjubScalar` perform modular reduction by the Jubjub scalar modulus,
- casts from `JubjubScalar` to `Field` always succeed (the maximum `Field` value is larger than the maximum `JubjubScalar` value), and
- casts from `JubjubScalar` to `Uint` are checked at runtime and will fail if the value is too large for the target `Uint` type (the largest `Uint` type is currently too small to hold the maximum `JubjubScalar`).

Equals and not-equals comparisons are supported for `JubjubScalar` values.  Arithmetic is not (currently) supported.

For consistency with `JubjubScalar` and for future "big" `Uint` types, this change **has a major breaking change** to the semantics of `Field` and `Uint` types.  `Field` is no longer a supertype of all `Uint` types, instead `Field` and `Uint` types are unrelated by subtyping.  There are still casts in both directions between `Field` and `Uint` types, with the same semantics as before.  The former "upcasts" from `Uint` to `Field` must now be explicit.

There are two awkward things that are not yet fixed in this change:

- Numeric literals still have exact `Uint` types, and they will no longer implicitly be upcast to `Field`.  This will require an `as Field` cast for literals that are used at `Field` type.  We might find a way to make this less burdensome on programmers.
- Mixed arithmetic of `Field` and `Uint` valued expressions still works exactly as before: there is an implicit cast (though no longer an upcast) of the `Uint` operand to `Field` type and then `Field` arithmetic is performed.  This is inconsistent and can leave programmers guessing about whether an explicit cast is needed or not.  We will very likely fix this in a later change.

In the ZKIR v2 backend, `JubjubScalar` is represented by native field values, the operations (`ecMul`, `ecMulGenerator`) will perform the modular reduction.  In the ZKIR v3 backend it is represented as a value in the range 0..MAX_JUBJUB_SCALAR, and ZKIR v3 uses a type `Scalar<Jubjub>` for these values.  The ZKIR v3 output of the compiler is well-typed.


kmillikin:
There is still some work remaining on this PR:

- [ ] bump the compiler version
- [x] create a test that will fail due to missing support in ZKIR

compiler/langs.ss
```diff
@@ -300,6 +300,11 @@
       (targ-size src nat)                    => nat
       (targ-type src type)                   => type
       )
+    (Field-Type (ftype)
+      (field-native)
+      (field-scalar ctype))
+    (Curve-Type (ctype)
```
dybvig, line 307: Can `ctype` just be a terminal defined by:
```
(curve-type (ctype))
```
and `curve-type?` is defined by:
```
(define (curve-type? x)
  (let ([curve-type* '(curve-jubjub)])
    (and (memq x curve-type*) #t)))
```

compiler/langs.ss
```diff
@@ -1115,25 +1144,27 @@
...
+         (== triv1 triv2)                            => (== triv1 3 triv2)
+         (select triv0 triv1 triv2)                  => (select triv0 triv1 triv2)
          (bytes-ref triv nat)
-         (bytes->field src len triv1 triv2)      => (bytes->field len #f triv1 #f triv2)
-         (vector->bytes triv triv* ...)          => (vector->bytes triv triv* ...) ; result holds one field's worth of bytes
-         (downcast-unsigned src safe (maybe nat?) nat triv) =>
-           (downcast-unsigned safe nat? nat triv)))
+         (bytes->field src len triv1 triv2)          => (bytes->field len #f triv1 #f triv2)
+         (vector->bytes triv triv* ...)              => (vector->bytes triv triv* ...) ; result holds one field's worth of bytes
+         (cast-to-field ftype primitive-type triv)
```
dybvig, line 1167: Does `cast-to-field` cast from a non-native field to a native field?  Can `ftype` ever be a native field here?
Similarly for `cast-from-field`.

compiler/langs.ss
dybvig: Or even simpler still, at least in terms of diffs:
```
(Type (type)
  ---
  (tfield src)             => (tfield)
  (tfield src ctype)       => (tfield ctype)
  ---)
```
where `ctype` is a terminal defined by:
```
(curve-type (ctype))
```
and `curve-type?` is defined by:
```
(define (curve-type? x)
  (let ([field-type* '(native curve-jubjub)])
    (and (memq x field-type*) #t)))
```

kmillikin: Yeah, it could be.  We will eventually have more field types: `Field` is `(field-native)` (and would also be `(field-scalar (curve-bls12-381))` and `(field-base (curve-jubjub))`, but the compiler will not generate those), `JubjubScalar` is `(field-scalar (curve-jubjub))`, `Secp256k1Base` is `(field-base (curve-secp256k1))`, `Secp256k1Scalar` is `(field-scalar (curve-sec256k1))`.

We will also eventually use `Curve-Type` for types like `(point (curve-secp256k1))` and `(point (curve-jubjub))` when I (or someone else) removes the "opaque" hack for representing those.

I also thought about enumerating in the representation just exactly which field types were supported, which might be better anyway.

The Secp256k1 change does have some cleanup of the use sites.

compiler/langs.ss
kmillikin: The are admittedly very warty right now: `cast-to-field` will cast in both directions between `Field` and `JubjubScalar`.  `cast-from-field` will cast from both field types to an unsigned integer type.  It's only here because we have to handle the `JubjubScalar` -> `Uint` cast differently in ZKIR v2 and v3 (and it's used in place of `downcast-unsigned` for `Field` -> `Uint` which isn't a downcast now, and mostly had two variants (from Field and from Uint) before anyway.

I think it's future work to fill out the available casts, which are ad hoc right now.  In keeping with what I wrote above, we might want the IR to only allow exactly the casts that we support.

kmillikin:
This was incorporated in #519

PR #184, opened 2026-03-03
Implement `default<Opaque<'JubjubPoint'>>` correctly in circuit

kmillikin:
In ZKIR v2, this needs to be represented as a _pair_ of field values (0,1), not a single (0) field value.

In ZKIR v3, this needs to be constructed via an operation.  We use `decode(Point<Jubjub>,0,1)`.

This is all implemented by adding `default` as an instruction in the language `Lflattened`, but only for ZKIR opaque types (currently only `JubjubPoint`). Since `default` is translated from `Lcircuit` to an `Lflattened` instruction, we no longer consider it trivial in `Lcircuit`.

### DIFF
--- a/compiler/test.ss
+++ b/compiler/test.ss
@@ -46073,6 +46073,40 @@ groups than for single tests.
           (= %t.1 (== %ix.0 0))
           (%t.1))))
     )
+  ;; Unused `default`.
+  (test
+    '(
+      "import CompactStandardLibrary;"
+      "ledger impure: Boolean;"
+      "export circuit fisk(pt: JubjubPoint): JubjubPoint {"
+      "  impure = true;"
+      "  const pt0 = default<JubjubPoint>;"
+      "  const pt1 = ecAdd(pt, pt);"
+      "  return pt1;"
+      "}"
+      )
+    (returns
+      (program
+        (kernel-declaration (%kernel.0 () (Kernel)))
+        (public-ledger-declaration
+          ((%impure.1
+             (0)
+             (__compact_Cell (ty ((abytes 1)) ((tfield 1)))))))
+        (native %ecAdd.2 ((argument
+                            (%a.3 %a.4)
+                            (ty ((afield) (afield)) ((tfield) (tfield))))
+                          (argument
+                            (%b.5 %b.6)
+                            (ty ((afield) (afield)) ((tfield) (tfield)))))
+             (ty ((afield) (afield)) ((tfield) (tfield))))
+        (circuit %fisk.7 ((argument
+                            (%pt.8 %pt.9)
+                            (ty ((afield) (afield)) ((tfield) (tfield)))))
+             (ty ((afield) (afield)) ((tfield) (tfield)))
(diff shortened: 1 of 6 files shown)


compiler/test.ss
```diff
@@ -38086,57 +38086,65 @@ groups than for single tests.
           (= %t.76 (public-ledger #t %field5.6 (5) resetToDefault))
           (= %t.77 (public-ledger #t %field6.7 (6) resetToDefault))
           (= %value.78 (public-ledger #t %kernel.0 () self))
-          (= %tmp.79
+          (= %t.79
```
kmillikin: This is named because `default` is no longer trivial.  Then the rest of the identifier indexes are off by one.

compiler/test.ss
```diff
@@ -57270,7 +57278,50 @@ groups than for single tests.
...
+        "    { \"op\": \"declare_pub_input\", \"var\": 3 },"
+        "    { \"op\": \"declare_pub_input\", \"var\": 0 },"
+        "    { \"op\": \"declare_pub_input\", \"var\": 0 },"
+        "    { \"op\": \"declare_pub_input\", \"var\": 0 },"
+        "    { \"op\": \"declare_pub_input\", \"var\": 0 },"
+        "    { \"op\": \"pi_skip\", \"guard\": 0, \"count\": 5 },"
+        "    { \"op\": \"load_imm\", \"imm\": \"91\" },"
+        "    { \"op\": \"declare_pub_input\", \"var\": 4 },"
+        "    { \"op\": \"pi_skip\", \"guard\": 0, \"count\": 1 },"
+        "    { \"op\": \"ec_add\", \"a_x\": 2, \"a_y\": 0, \"b_x\": 2, \"b_y\": 0 },"
```
kmillikin: `default<'JubjubPoint'>` consists of the pair of instructions, one with index 2 (`load_imm 0x0`) and one with index 0 (`load_imm 0x1`).

compiler/test.ss
```diff
@@ -63614,6 +63665,32 @@ groups than for single tests.
...
+      '(
+        "{"
+        "  \"version\": { \"major\": 3, \"minor\": 0 },"
+        "  \"do_communications_commitment\": false,"
+        "  \"inputs\": ["
+        "  ],"
+        "  \"instructions\": ["
+        "    { \"op\": \"impact\", \"guard\": \"0x01\", \"inputs\": [\"0x10\", \"0x01\", \"0x01\", \"0x01\", \"0x00\", \"0x11\", \"0x01\", \"0x01\", \"0x01\", \"0x01\", \"0x91\"] },"
+        "    { \"op\": \"decode\", \"type\": \"Point<Jubjub>\", \"output\": \"%pt.0\", \"inputs\": [\"0x00\", \"0x01\"] },"
+        "    { \"op\": \"add\", \"output\": \"%t.1\", \"a\": \"%pt.0\", \"b\": \"%pt.0\" },"
```
kmillikin: `default<'JubjubPoint'>` is the result of `decode Point<Jubjub> 0 1`.

PR #256, opened 2026-03-25
perf: preserve structural sharing for nested ledger ADT types

RomarQ:
Related to https://github.com/LFDT-Minokawa/compact/pull/255, instead of just adding a guard, it adds an optimization to preserve structurally shared `tadt` nodes across nanopass passes.

## Problem

The Compact compiler crashes with OOM when compiling contracts with deeply nested `Map` types (e.g. `Map<Field, Map<Field, Map<...>>>`). Depth 4 compiles fine, but depth 7 causes OOM.

Each nanopass pass auto-transforms through `tadt` nodes, rebuilding the entire subtree even when nothing changes. The `expand-modules-and-types` pass creates structurally shared `tadt` nodes (multiple references point to the same Scheme object), but every subsequent pass breaks this sharing by creating independent copies. With ~14 passes in the analysis pipeline each independently expanding the tree, memory grows exponentially.

Measured before this fix:
- Depth 6: 3.4 GB, 10s
- Depth 8: OOM

## Fix

Add explicit `tadt` clauses to the analysis passes so they preserve structural sharing instead of blindly rebuilding:

- **Cross-language passes** use an `eq?`-keyed hash table so shared input nodes produce shared output nodes.
- **Same-language passes** return the original node unchanged, since they don't modify types.

Measured after this fix:
- Depth 6: 80 MB, 0.05s
- Depth 8: 80 MB, 0.05s
- Depth 16: 80 MB, 0.06s


RomarQ:
@kmillikin Did you have a look at this issue? I think the fix should be simple

As a simple mitigation there is also this PR that just adds an ADT nesting limit: https://github.com/LFDT-Minokawa/compact/pull/255

PR #274, opened 2026-03-30
Add `keccak256` to the standard library

kmillikin:
The signature is the same as for `persistentHash` (SHA-256).  It is implemented in the JS target by a runtime function `keccak256` with the same signature as the existing runtime function `persistentHash`.  It is implemented in ZKIR v3 with a single new operation `keccak256` with the same signature as the existing `persistent_hash` operation.

It is unsupported in ZKIR v2 and it is a compile-time error to use it in an impure circuit without passing the experimental feature flag `--feature-zkir-v3`.


dybvig (approved):

kmillikin:
@golddydev will add tests in a separate PR.

PR #276, opened 2026-03-31
Workaround for unconditionally failing conditional constraints

dybvig:
Addresses Issue 226 with temporary compiler workarounds.  See the issue and the change-log entry for details.

To aid reviewing, I have created a separate branch (`every-assignment-guard`) that moves the tests (conditional guards) from assignment statement right-hand sides to the assignment statements themselves.  I suggest [comparing `every-assignment-guard` with `main`](https://github.com/LFDT-Minokawa/compact/compare/main...every-assignment-guard) then [comparing this branch (`issue-226-conditional-constraints`) with `every-assignment-guard`](https://github.com/LFDT-Minokawa/compact/compare/every-assignment-guard...issue-226-conditional-constraints).

### DIFF
--- a/compiler/circuit-passes.ss
+++ b/compiler/circuit-passes.ss
@@ -1946,7 +1946,7 @@
             (cons*
               ; t1 = triv && test
               `(= (quote #t) ,t1 (select ,triv ,test (quote #f)))
-              ; t1 = !triv && test
+              ; t2 = !triv && test
               `(= (quote #t) ,t2 (select ,triv (quote #f) ,test))
               (k t1 t2)))))
       )
@@ -3234,31 +3234,31 @@
     )
 
   (define-pass missing-guard-workarounds : Lflattened (ir) -> Lflattened ()
-    ; this pass implements workarounds for the lack of conditionality of
-    ; certain zkir operators.  the lack of conditionality burns in one of
+    ; This pass implements workarounds for the lack of conditionality of
+    ; certain zkir operators.  The lack of conditionality burns in one of
     ; two ways: explicit checks like constrain_bits fail even when the
     ; conditional says not to execute it, and implicit operand checks, e.g.,
     ; by less_than, fail because an input value is undefined and might have
     ; any value due to the conditionality of the input value's computation.
-    ; to avoid being overly paranoid, the pass records whether a variable
-    ; definitly has a value and skips remediation for unknown values when
-    ; a variable is defined.  it also implements various special cases to
+    ; To avoid being overly paranoid, the pass records whether a variable
+    ; definitely has a value and skips remediation for unknown values when
+    ; a variable is defined.  It also implements various special cases to
     ; avoid generating the worst-case code unless necessary.
     ;
-    ; once zkir implements conditionality for the operators that can fail,
+    ; Once zkir implements conditionality for the operators that can fail,
     ; this pass can simply be removed.
     (definitions
       (define-syntax with-temp-ids
         (syntax-rules ()
           [(_ src (t ...) b1 b2 ...)
            (let* ([t (make-temp-id src 't)] ...) b1 b2 ...)]))
-      (module (defined! defined?)
-        (define def-ht (make-eq-hashtable))
+      (module (def-ht defined! defined?)
+        (define def-ht)
         (define (defined! var-name) (hashtable-set! def-ht var-name #t))
         (define (defined? triv)
           (or (not (id? triv))
               (hashtable-contains? def-ht triv))))
-      (define (insure-defined src test triv k)
+      (define (ensure-defined src test triv k)
         (if (defined? triv)
             (k triv)
             (with-output-language (Lflattened Statement)
@@ -3267,8 +3267,14 @@
                       (k t)))))))
     (Circuit-Definition : Circuit-Definition (ir) -> Circuit-Definition ()
       [(circuit ,src ,function-name (,arg* ...) ,type ,stmt* ... (,triv* ...))
-       (let ([stmt* (apply append (maplr Statement stmt*))])
-         `(circuit ,src ,function-name (,arg* ...) ,type ,stmt* ... (,triv* ...)))])
+       (fluid-let ([def-ht (make-eq-hashtable)])
+         (for-each
+           (lambda (arg)
+             (nanopass-case (Lflattened Argument) arg
+               [(argument (,var-name* ...) ,type) (for-each defined! var-name*)]))
+           arg*)
+         (let ([stmt* (apply append (maplr Statement stmt*))])
+           `(circuit ,src ,function-name (,arg* ...) ,type ,stmt* ... (,triv* ...))))])
     (Statement : Statement (ir) -> * (stmt*)
       [(= ,test ,var-name ,single)
        (when (eqv? test 1) (defined! var-name))
@@ -3297,9 +3303,9 @@
     (Single : Single (ir test var-name) -> * (stmt*)
       [(< ,bits ,triv1 ,triv2)
        (with-output-language (Lflattened Statement)
-         (insure-defined (id-src var-name) test triv1
+         (ensure-defined (id-src var-name) test triv1
            (lambda (triv1)
-             (insure-defined (id-src var-name) test triv2
+             (ensure-defined (id-src var-name) test triv2
                (lambda (triv2)
                  (list `(= 1 ,var-name (< ,bits ,triv1 ,triv2))))))))]
       [(bytes->field ,src ,len ,triv1 ,triv2)
@@ -3314,9 +3320,9 @@
              ; less than or equal to q, and when triv1 = q, triv2 must be less
              ; than or equal to r.
              (let-values ([(q r) (div-and-mod (max-field) (expt 256 (field-bytes)))])
-               (insure-defined (id-src var-name) test triv1
+               (ensure-defined (id-src var-name) test triv1
                  (lambda (triv1)
-                    (insure-defined (id-src var-name) test triv2
+                    (ensure-defined (id-src var-name) test triv2
                       (lambda (triv2)
                         (with-temp-ids (id-src var-name) (t1 t2 t3 t4 t5 t6 t7)
                           (list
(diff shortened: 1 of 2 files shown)


compiler/zkir-passes.ss
```diff
@@ -347,9 +344,21 @@
                                   [(ty (,alignment* ...) (,primitive-type* ...)) (length primitive-type*)]))])
              (list ledger-op (apply + (type-length type) (map type-length type*))))])
         (Statement : Statement (ir) -> * (void)
+          [(= ,var-name (downcast-unsigned ,src ,safe ,[* test] ,nat ,[* triv]))
+           (unless safe
+             (constrain-type (with-output-language (Lflattened Primitive-Type)
+                                                   `(tfield ,nat))
+                             triv))
+           (if (id? triv)
```
kmillikin: This pass is really hard to reason about!

Isn't `triv` here always a memory index (a ZKIR instruction output number)?  I also don't immediately see why this instruction ignores `test`.

dybvig: You're right, `Triv` always returns an index, and the first case was never taken.  Now there is only one case.

compiler/circuit-passes.ss
```diff
@@ -2486,15 +2522,39 @@
...
-                          `(assert ,src ,var-name^ "bytes value is too big to fit in a field")
+                          `(= ,t1 (== ,triv 0))
+                          `(= ,t2 (select ,test ,t1 1)) 
+                          `(assert ,src ,t2 "bytes value is too big to fit in a field")
                           ls))))
-                  (let-values ([(triv1 triv2) (apply values (list-tail triv* n))])
-                    (with-output-language (Lflattened Statement)
-                      (list `(= ,var-name (bytes->field ,src ,test ,len ,triv1 ,triv2)))))
+                  (if (eqv? test 1)
+                      (let-values ([(triv1 triv2) (apply values (list-tail triv* n))])
```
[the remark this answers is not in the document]
dybvig: As you can see, this pass no longer does the workarounds.  They are all done by the new pass `missing-guard-workarounds`, and I don't think this problem exists there.

compiler/circuit-passes.ss
```diff
@@ -2486,15 +2522,39 @@
...
+                        (with-output-language (Lflattened Statement)
+                          (let ([t1 (make-temp-id src 't1)]
+                                [t2 (make-temp-id src 't2)]
+                                [t3 (make-temp-id src 't3)]
+                                [t4 (make-temp-id src 't4)]
+                                [t5 (make-temp-id src 't5)]
+                                [t6 (make-temp-id src 't6)]
+                                [t7 (make-temp-id src 't7)])
+                            (let-values ([(q r) (div-and-mod (max-field) (expt 256 (field-bytes)))])
```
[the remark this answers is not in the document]
dybvig: I tried to explain this in the new pass.

compiler/zkir-passes.ss
dybvig: I don't understand your question about `test`.  The point of the workarounds is that `zkir` doesn't yet have the guards that would let the code use `test` in any useful way.

dybvig:
Handling the workarounds in `flatten-datatypes` is too unwieldy and doesn't allow for easy removal of the workarounds when they are no longer needed, so I have added a new pass `missing-guard-workarounds` that does the removal.  The new pass fixes some unknown-value issues with the `flatten-datatypes` workarounds, avoids the use of `div-mod-power-of-two` for all unsigned-to-unsigned casts, and tracks when variables are definitely defined to avoid unknown-value workarounds for references to those values.  `downcast-unsigned` now sports a `(maybe nat?)` field that is false when the cast-from type is `Field` and the maximum cast-from value when the cast-from type is a `Uint`.  Every assignment form in `Lcircuit` and `Lflattened` includes a `test` (guard) field, and the `test` fields have been removed from the `Rhs` forms.  This is as it used to be and is necessary for tracking when variables are definitely defined.  It will also make possible the use of the missing guards once they are supported by `zkir` if we don't switch to a different, e.g., nested, structure altogether for the input to `zkir`.

compiler/zkir-passes.ss
kmillikin: Never mind the question about `test`, I wrote that hastily and my thinking was wrong.

compiler/test.ss
```diff
@@ -52066,10 +52094,12 @@ groups than for single tests.
...
         "    { \"op\": \"load_imm\", \"imm\": \"0004\" },"
         "    { \"op\": \"add\", \"a\": 1, \"b\": 7 },"
-        "    { \"op\": \"constrain_bits\", \"var\": 8, \"bits\": 10 },"
-        "    { \"op\": \"copy\", \"var\": 8 },"
-        "    { \"op\": \"cond_select\", \"bit\": 0, \"a\": 9, \"b\": 4 },"
-        "    { \"op\": \"output\", \"var\": 10 }"
+        "    { \"op\": \"cond_select\", \"bit\": 0, \"a\": 8, \"b\": 4 },"
+        "    { \"op\": \"less_than\", \"a\": 9, \"b\": 7, \"bits\": 11 },"
+        "    { \"op\": \"cond_select\", \"bit\": 0, \"a\": 10, \"b\": 2 },"
+        "    { \"op\": \"assert\", \"cond\": 11 },"
```
kmillikin, line 52096: This is not necessarily something we need to fix in this PR, because I don't think it's a correctness issue.

But can't we just directly `assert(10)` (that is, assert the result of the `less_than` comparison) because we've ensured that it's true when the branch is not taken?  The `cond_select(0, 10, 2)` (2 is `load_imm(0x1)`) looks unnecessary.

compiler/test.ss
```diff
@@ -52066,10 +52094,12 @@ groups than for single tests.
...
+        "    { \"op\": \"cond_select\", \"bit\": 0, \"a\": 8, \"b\": 4 },"
+        "    { \"op\": \"output\", \"var\": 12 }"
```
kmillikin, line 52098: Instruction output 12 is the same as instruction output 9, so I think this `cond_select` just above is unnecessary.

compiler/test.ss
```diff
@@ -53318,14 +53348,15 @@ groups than for single tests.
...
         "    { \"op\": \"add\", \"a\": 3, \"b\": 5 },"
-        "    { \"op\": \"less_than\", \"a\": 3, \"b\": 5, \"bits\": 12 },"
-        "    { \"op\": \"cond_select\", \"bit\": 11, \"a\": 7, \"b\": 5 },"
-        "    { \"op\": \"cond_select\", \"bit\": 2, \"a\": 5, \"b\": 12 },"
-        "    { \"op\": \"assert\", \"cond\": 13 },"
+        "    { \"op\": \"cond_select\", \"bit\": 2, \"a\": 7, \"b\": 3 },"
+        "    { \"op\": \"less_than\", \"a\": 11, \"b\": 5, \"bits\": 12 },"
+        "    { \"op\": \"cond_select\", \"bit\": 12, \"a\": 7, \"b\": 5 },"
+        "    { \"op\": \"cond_select\", \"bit\": 2, \"a\": 5, \"b\": 13 },"
+        "    { \"op\": \"assert\", \"cond\": 14 },"
```
kmillikin: The difference here is the `less_than` comparison.  I guess we guard the first input because the operation can fail (if both inputs don't fit in the number of bits)?  Here we *should* know that it can't fail, the first input (originally 3) is already constrained to fit in 12 bits.

compiler/circuit-passes.ss
```diff
@@ -1938,8 +1944,10 @@
         (let ([t1 (make-temp-id src 't)] [t2 (make-temp-id src 't)])
           (with-output-language (Lcircuit Statement)
             (cons*
-              `(= ,t1 (select ,triv ,test (quote #f)))
-              `(= ,t2 (select ,triv (quote #f) ,test))
+              ; t1 = triv && test
+              `(= (quote #t) ,t1 (select ,triv ,test (quote #f)))
+              ; t1 = !triv && test
```
kmillikin: t1 ==> t2 (in the comment)

compiler/circuit-passes.ss
```diff
@@ -3194,6 +3233,171 @@
       [else ir])
     )
 
+  (define-pass missing-guard-workarounds : Lflattened (ir) -> Lflattened ()
+    ; this pass implements workarounds for the lack of conditionality of
```
kmillikin: Capitalize please :)

compiler/circuit-passes.ss
```diff
@@ -3194,6 +3233,171 @@
...
+    ; certain zkir operators.  the lack of conditionality burns in one of
+    ; two ways: explicit checks like constrain_bits fail even when the
+    ; conditional says not to execute it, and implicit operand checks, e.g.,
+    ; by less_than, fail because an input value is undefined and might have
+    ; any value due to the conditionality of the input value's computation.
+    ; to avoid being overly paranoid, the pass records whether a variable
+    ; definitly has a value and skips remediation for unknown values when
```
kmillikin: definitly ==> definitely

compiler/circuit-passes.ss
```diff
@@ -3194,6 +3233,171 @@
...
+        (syntax-rules ()
+          [(_ src (t ...) b1 b2 ...)
+           (let* ([t (make-temp-id src 't)] ...) b1 b2 ...)]))
+      (module (defined! defined?)
+        (define def-ht (make-eq-hashtable))
+        (define (defined! var-name) (hashtable-set! def-ht var-name #t))
+        (define (defined? triv)
+          (or (not (id? triv))
+              (hashtable-contains? def-ht triv))))
+      (define (insure-defined src test triv k)
```
kmillikin: Should be `ensure-defined`?

compiler/circuit-passes.ss
```diff
@@ -3194,6 +3233,171 @@
...
+        (if (defined? triv)
+            (k triv)
+            (with-output-language (Lflattened Statement)
+              (with-temp-ids src (t)
+                (cons `(= 1 ,t (select ,test ,triv 0))
+                      (k t)))))))
+    (Circuit-Definition : Circuit-Definition (ir) -> Circuit-Definition ()
+      [(circuit ,src ,function-name (,arg* ...) ,type ,stmt* ... (,triv* ...))
+       (let ([stmt* (apply append (maplr Statement stmt*))])
+         `(circuit ,src ,function-name (,arg* ...) ,type ,stmt* ... (,triv* ...)))])
```
kmillikin: I think the arguments are always defined.

dybvig: Great catch!  I had a `for-each` over the arguments in a separate temporary branch but somehow lost it in the merge.

compiler/circuit-passes.ss
dybvig: It's murky, but I also like ensure better here.

PR #277, opened 2026-03-31
Update: giles-dependabot-defaults

gilescope:
set safer defaults for dependabot but also turn on dependabot for cargo, npm as they're used in this repo. I'm unclear what we do about the secret?


JosephDenman (approved):
LGTM!

Issue #183, opened 2026-03-02
Expose Poseidon Hashing Algorithm From Compact Lang


emnul:
## Describe the proposed feature:
At this time, there is no circuit efficient hashing algorithm available in Compact that can be used to derive state data. Developers must use the `persistentHash` circuit which leads to very large circuits / proving times. Exposing the Poseidon hash algorithm as a standalone API would enable developers to derive state data using an efficient hashing algorithm.

## User Need:
Many of our contracts derive state using `persistentHash` causing them to be very large. Complex systems will not be able to utilize our contracts as the circuit size would exceed allowable deployment size.

## Expected Benefit:
A stable API for the Poseidon hashing algorithm would enable developers to create more efficient circuits.

kmillikin:
`transientHash` is (currently) Poseidon.

dybvig:
Also SHA which is persistentHash

kmillikin:
The names we had settled on when we discussed this were `poseidon` and `sha256` (to go with `keccak256`).

The original hashing circuits have `xCommit` versions in the standard library (and we implement them as natives for some reason?).  It's easy enough to implement them in Compact, I think:

```
circuit xCommit<T>(value: T, rand: RandType): ResultType {
  return xHash<[RandType, T]>([rand, value]);
}
```

Where `RandType` is whatever you want it to be and `ResultType` is the same as the return type of `xHash`.  Maybe we should find a way to get that into the standard library docs.

dybvig:
These should first be appropriately renamed in the on-chain runtime and this change can be propagated through to the compact standard library, which can then publish the OCRT `poseidon` as both `poseidon` and `transientHash`.  To rename only in the standard library would risk an accidental switch of what the standard library calls `poseidon` to something that isn't poseidon if the implementation of the OCRT `transientHash` function changes.

PR #308, opened 2026-04-06
Implement Schnorr signatures over Jubjub in compact standard library

### LINKED ISSUE #163: Add Schnorr signature over JubJub in Standard library
Add native Schnorr signature support over the JubJub curve into the compact standard library. This enhancement implements both signing and verification primitives directly in the language’s core library, allowing compact programs to leverage Schnorr cryptographic signatures without depending on external C/Rust bindings. Work has already started in https://github.com/LFDT-Minokawa/compact/pull/65  

## User Need:
Developers writing cryptographically secure compact programs need first-class support for Schnorr signatures on JubJub. Use cases include:
	•	Secure message authentication in on-chain or off-chain compact logic.
	•	Signature verification workflows within compact smart contracts or embedded logic.
	•	Native support without external dependencies, making standard library code safer, easier to audit, and more performant.
	•	Testable and maintainable cryptographic primitives, with documentation and tests included.

This feature ensures that cryptographic use cases are accessible out of the box, meeting common zero-knowledge and signature verification patterns without requiring bespoke external bindings.

## Expected Benefit:
	•	Expanded language crypto capability: Users can write and verify Schnorr-signed messages directly in compact with native constructs.
	•	Performance and security improvements: Native implementation minimizes FFI overhead and surface area for memory boundary issues.
	•	Better developer experience: Maintainers and users can rely on standard library docs, tests, and tooling rather than custom solutions.
	•	Broader applicability: Compact programs can now participate in ecosystems and protocols that standardize on Schnorr signatures (e.g., ZK proofs, multi-signatures).

Overall, this enhances compact as a secure, expressive language for cryptographic applications.

iquerejeta:
Closes #163 . Originally implemented here https://github.com/LFDT-Minokawa/compact/pull/65

This PR should be merged once https://github.com/midnightntwrk/midnight-ledger/pull/165 lands. 

Reviewers should pay particular attention to the TypeScript implementation of the Schnorr primitives, as this part of the code was developed with the assistance of Claude. I've added tests to ensure the logic matches the in-circuit verifier. However, its the first time I touch TS code, so it should be reviewed with care.


runtime/src/built-ins.ts
```diff
@@ -232,3 +232,100 @@ export function alignedConcat(...values: ocrt.AlignedValue[]): ocrt.AlignedValue
...
+    ...CompactTypeJubjubPoint.alignment(),
+    ...CompactTypeJubjubPoint.alignment(),
+    ...rtType.alignment(),
+  ];
+  const challengeValue: ocrt.Value = [
+    ...CompactTypeJubjubPoint.toValue(announcement),
+    ...CompactTypeJubjubPoint.toValue(pk),
+    ...rtType.toValue(msg),
+  ];
+  const c = reduceModJubjubOrder(ocrt.valueToBigInt(ocrt.transientHash(challengeAlignment, challengeValue)));
```
SupernaviX, line 343: `transientHash` is [not documented as Poseidon](https://docs.midnight.network/api-reference/compact-runtime/functions/transientHash). The impl [is Poseidon](https://github.com/midnightntwrk/midnight-ledger/blob/909292455347b9bdbe6476cddd1fc9b09e78fd70/transient-crypto/src/hash.rs#L75-L81), but is documented as something that can change across hard forks.

I think this should use an explicitly stable hash function (I don't mean `persistentHash`, I just mean "something with the current impl of `transientHash` but guaranteed not to change")

iquerejeta: Agreed - There is the plan to add explicitly names hash functions in compact (see [here](https://github.com/LFDT-Minokawa/compact/issues/183#issuecomment-3991453870)), and once we have that, instead of using the 'transientHash' wrapper, we'll directly call Poseidon.

runtime/src/built-ins.ts
```diff
@@ -232,3 +232,100 @@ export function alignedConcat(...values: ocrt.AlignedValue[]): ocrt.AlignedValue
   }
   return res;
 }
+
+/**
+ * Samples a random JubJub scalar.
+ *
+ * The returned value is in the range [0, JUBJUB_SCALAR_ORDER).
+ */
+export function jubjubSampleScalar(): bigint {
```
JosephDenman, line 261: It looks like this function is used to generate secret keys. I would consider adding an alias for it like`sampleJubjubSchnorrSk` .

compiler/standard-library.compact
```diff
@@ -350,3 +350,36 @@ export circuit unshieldedBalanceGt(color: Bytes<32>, amount: Uint<128>): Boolean
...
+  ann_y: Field,
+  pk_x: Field,
+  pk_y: Field,
+  msg: Vector<n, Field>
+}
+
+// Verify Schnorr signature over the native curve with a message of size #n.
+export circuit jubjubSchnorrVerify<#n>(msg: Vector<n, Field>, signature: JubjubSchnorrSignature, pk: JubjubPoint): [] {
+  const {announcement, response} = signature;
+  const c_native: Field = transientHash<JubjubSchnorrHashInput<n>>(JubjubSchnorrHashInput<n>{
```
JosephDenman: `cNative` instead of `c_native`.

compiler/standard-library.compact
```diff
@@ -350,3 +350,36 @@ export circuit unshieldedBalanceGt(color: Bytes<32>, amount: Uint<128>): Boolean
...
+struct JubjubSchnorrHashInput<#n> {
+  ann_x: Field,
+  ann_y: Field,
+  pk_x: Field,
+  pk_y: Field,
+  msg: Vector<n, Field>
+}
+
+// Verify Schnorr signature over the native curve with a message of size #n.
+export circuit jubjubSchnorrVerify<#n>(msg: Vector<n, Field>, signature: JubjubSchnorrSignature, pk: JubjubPoint): [] {
```
JosephDenman: So, we don't have a way to catch and recover from errors in Compact at the moment. With an `assert` in return position, contract's can't have conditional logic that allows them to do something when a signature is invalid. I would suggest having this function return a boolean.

test-center/ts/schnorr.ts
```diff
@@ -0,0 +1,67 @@
...
+
+test('TypeScript signature passes in-circuit verification', () => {
+  const sk = runtime.jubjubSampleScalar();
+  const pk = runtime.jubjubSchnorrVerifyingKey(sk);
+  const msg = sampleMsg();
+
+  const sig = runtime.jubjubSchnorrSign(msgType, msg, sk);
+
+  const [c, Ctxt] = startContract(contractCode, {}, 0);
+  expect(() => c.circuits.verifySchnorrN3(Ctxt, msg, sig, pk)).not.toThrow();
```
JosephDenman: Probably we want `verifySchnorrN3` to return a boolean and have these tests assert that the correct `true`/`false` value is returned instead of just asserting that an error isn't thrown. This pairs with my suggestion that `jubjubSchnorrVerify` returns a boolean.

runtime/src/built-ins.ts
```diff
@@ -232,3 +232,100 @@ export function alignedConcat(...values: ocrt.AlignedValue[]): ocrt.AlignedValue
...
+export function reduceModJubjubOrder(value: bigint): bigint {
+  return value % JUBJUB_SCALAR_ORDER;
+}
+
+/**
+ * Derives the Schnorr verifying key (public key) from a signing key.
+ *
+ * Equivalent to {@link ecMulGenerator}(sk).
+ */
+export function jubjubSchnorrVerifyingKey(sk: bigint): JubjubPoint {
```
JosephDenman: We should pick a convention and use it uniformly - either "signing key" and "verifying key" or "sk" and "pk". The `onchain-runtime-wasm` package uses "signingKey" explicitly, so maybe we should adopt that?

What I mean is, for example, since you're calling the argument to `jubjubSchnorrVerifyingKey` "`sk`", maybe the function should be named `jubjubSchnorrPk`, or call the argument `signingKey`.

test-center/ts/schnorr.ts
```diff
@@ -0,0 +1,67 @@
...
+// NOTE: this file is included as a test body inside a describe block by
+// compiler/test.ss; it therefore has access to `contractCode`, `startContract`,
+// `expect`, `test`, and `runtime` (= @midnight-ntwrk/compact-runtime) without
+// explicit imports.
+
+const MSG_LEN = 3;
+const msgType = new runtime.CompactTypeVector(MSG_LEN, runtime.CompactTypeField);
+const sampleMsg = (): bigint[] => [1n, 2n, 3n];
+
+test('TypeScript Schnorr sign/verify roundtrip', () => {
```
JosephDenman: These first two tests are limited to tests of the function exposed through `@midnight-ntwrk/compact-runtime`, so the first two tests should go in `runtime/test`. The last two tests use a contract, so they can stay in `test-center`.

doc/api/CompactStandardLibrary/exports.md
```diff
@@ -41,6 +41,18 @@ struct NativePoint {
 }
 ```
 
+### `JubjubSchnorrSignature`
```
JosephDenman, line 44: Please also update `doc/api/CompactStandardLibrary/README.md`, probably with a new section on cryptographic signatures since we'll be doing RSA and ECDSA as well.

JosephDenman (changes requested):
The implementation itself looks solid. Exactly how I would've done it. Mostly feedback on code structure.

runtime/src/constants.ts
```diff
@@ -22,6 +22,11 @@ import * as ocrt from '@midnight-ntwrk/onchain-runtime-v3';
  */
 export const MAX_FIELD: bigint = ocrt.maxField();
 
+/**
+ * The order of the JubJub scalar field.
+ */
+export const JUBJUB_SCALAR_ORDER: bigint = 0xe7db4ea6533afa906673b0101343b00a6682093ccc81082d0970e5ed6f72cb7n;
```
kmillikin: This is inconsistent with the way that we currently handle the native field type.  For native fields:

- We have `MAX_FIELD` (just above) which is read from the on-chain runtime, not a `bigint` literal and differs from the field order by 1
- We have `FIELD_MODULUS` ("modulus", not "order") in `built-ins.ts`, which is computed from `MAX_FIELD`

I honestly don't have a strong opinion about which is better, but we should be consistent.  What do you think @iquerejeta and @JosephDenman ?

runtime/src/built-ins.ts
```diff
@@ -232,3 +232,105 @@ export function alignedConcat(...values: ocrt.AlignedValue[]): ocrt.AlignedValue
...
+
+/**
+ * Alias for {@link jubjubSampleScalar}. Samples a random JubJub Schnorr signing key.
+ */
+export const sampleJubjubSchnorrSk = jubjubSampleScalar;
+
+/**
+ * Converts a native field element to a JubJub scalar by reducing modulo the scalar field order.
+ */
+export function jubjubScalarFromNative(x: bigint): bigint {
```
kmillikin: We have a file `casts.ts` that has some type casts but not all.  I honestly wonder how we decide which casts go where.

In the `JubjubScalar` foreign field PR I'm working on, I put the equivalent to this in `casts.ts`, implemented it in pure JS, and actually called it `convertNumericToJubjubScalar` because (1) the naming matches the other cast functions in that file `convertXxxToYyy` and (2) it works and is used to convert `Uint` types as well as native `Field`.

runtime/src/built-ins.ts
```diff
@@ -232,3 +232,105 @@ export function alignedConcat(...values: ocrt.AlignedValue[]): ocrt.AlignedValue
...
+  return ocrt.valueToBigInt(ocrt.jubjubScalarFromNative(ocrt.bigIntToValue(x)));
+}
+
+/**
+ * Reduce modulo the JubJub scalar field order.
+ *
+ * The returned value is in the range [0, JUBJUB_SCALAR_ORDER).
+ */
+export function reduceModJubjubOrder(value: bigint): bigint {
```
kmillikin, line 275: Isn't this exactly the same function as the one just above, or am I missing something?

runtime/src/built-ins.ts
```diff
@@ -232,3 +232,105 @@ export function alignedConcat(...values: ocrt.AlignedValue[]): ocrt.AlignedValue
...
+  return value % JUBJUB_SCALAR_ORDER;
+}
+
+/**
+ * Derives the Schnorr verifying key (public key) from a signing key.
+ *
+ * Equivalent to {@link ecMulGenerator}(signingKey).
+ */
+export function jubjubSchnorrVerifyingKey(signingKey: bigint): JubjubPoint {
+  return ecMulGenerator(signingKey);
```
kmillikin: Is this correct?

Do we need a call to `jubjubScalarFromNative(signingKey)` or `reduceModJubjubOrder(signingKey)` here?  Or do we need to put such a call in `ecMulGenerator` on the argument value?

compiler/midnight-natives.ss
```diff
@@ -83,6 +83,11 @@
    [y Field (discloses "a JubjubPoint containing y coordinate")])
   (TypeRef JubjubPoint))
 
+(declare-native-entry circuit jubjubScalarFromNative
```
kmillikin, line 95: A declaration like this in this file will:

1. inject this circuit `jubjubScalarFromNative` into the standard library, which seems like what you intend (at least, until we have a real `JubjubScalar` type and we can write `f as JubjubScalar`),
2. give JS function to use for it (`__compactRuntime.jubjubScalarFromNative`), and
3. expect to be handled explicitly by the compiler in the ZKIR code generator (that seems to be missing here).

Take a look at the equivalents for `constructJubjubPoint` just above:

- here: https://github.com/LFDT-Minokawa/compact/blob/main/compiler/zkir-passes.ss#L205
- and here: https://github.com/LFDT-Minokawa/compact/blob/main/compiler/zkir-v3-passes.ss#L90

kmillikin (commented):
Here are a few things I noticed.

runtime/src/constants.ts
JosephDenman: I agree that consistent naming conventions should be used. 

On the question of where the value lives, we could make an argument that the on-chain runtime doesn't actually need to know about or be responsible for housing information about Jubjub scalars. It does have a unique need to know about native fields.

iquerejeta: The JubJub scalar order is a fixed mathematical curve constant, unlike MAX_FIELD which may vary by proof system config. I'd say that these two are different, so I wouldn't be against the idea of keeping the literal here. 

As for the naming, I agree we should be consistent. I'll change to MODULUS once we agree on the other point.

runtime/src/built-ins.ts
iquerejeta: I'm happy to wait until that PR lands, and remove that function from here to directly use your function.

runtime/src/built-ins.ts
iquerejeta: They are indeed the same. I'll remove the `jubjubScalarFromNative` function, as the name of this one is more explicit with what it does.

runtime/src/built-ins.ts
iquerejeta: `signingKey` comes from `jubjubSampleScalar()` which already returns a canonical value, therefore the reduction here is not necessary.

compiler/midnight-natives.ss
iquerejeta: I'm not sure I understand this issue. One thing to keep in mind is that the in-circuit reduction happens implicitly. i.e., if you pass a value that is larger than the jubjub modulo it'll still pass. Does that answer your concern?

kmillikin: We still need code to tell the compiler how to generate ZKIR.  This program will fail to compile because of the missing ZKIR support:

```

ledger impure: Boolean;

export circuit test(): JubjubPoint {
  impure = true;
  return ecMulGenerator(jubjubScalarFromNative(0));
}
```

For ZKIR v2:

```
$ PATH=result/bin:$PATH compactc ~/Programs/compact/jubjub.compact /tmp/jubjub.out
PATH=result/bin:$PATH compactc ~/Programs/compact/jubjub.compact /tmp/jubjub.out
Exception: jubjub.compact line 7 char 25:
  unrecognized native circuit jubjubScalarFromNative
  ```

and for ZKIR v3:

```
$ PATH=result/bin:$PATH compactc --feature-zkir-v3 ~/Programs/compact/jubjub.compact /tmp/jubjub.out
PATH=result/bin:$PATH compactc --feature-zkir-v3 ~/Programs/compact/jubjub.compact /tmp/jubjub.out
unknown native: jubjubScalarFromNative
Internal error (please report): Exception: failed assertion not-implemented at line 147, char 25 of compiler/zkir-v3-passes.ss
```

If it's truly a no-op because the ZKIR implementation will perform the modular reduction, then I think it's pretty easy.  For ZKIR v2 in `zkir-passes.ss` we need a handler like:

```
(register-handler! 'jubjubScalarFromNative
  (lambda (align res* a)
    (bind-var! (car res*) a)))
```

And for ZKIR v3 in `zkir-v3-passes.ss` we need a handler like:

```
[(jubjubScalarFromNative)
 (assert (= (length var-name*) 1))
 (cons `(copy ,(car var-name*) ,(car triv*)) instr*)]
```

But I haven't tested that code.

Comment (kmillikin) [reply compiler/midnight-natives.ss:95]: To be clear: the circuits named in this file are the ones that have custom handling in the compiler for the ZKIR backend.  If you add such a circuit, you also have to add that custom handling to the compiler.

File: runtime/src/built-ins.ts (hunk 12 above)
Comment (kmillikin) [reply runtime/src/built-ins.ts]: This function is exported from the Compact runtime.  A user can directly call it from their DApp and pass any `bigint` as argument.

File: runtime/src/built-ins.ts (hunk 10 above)
Comment (kmillikin) [reply runtime/src/built-ins.ts]: We can land them in any order, it's just that one or the other of us will have to make a small change :)

File: runtime/src/constants.ts (hunk 9 above)
Comment (kmillikin) [reply runtime/src/constants.ts]: I'm OK with having the constant here as long as we can justify why we did it that way, and it sounds like we have at least two justifications.

File: compiler/midnight-natives.ss (hunk 13 above)
Comment (iquerejeta) [reply compiler/midnight-natives.ss:95]: Ok - I've added some changes to the schema files. Tests are passing, but the work is AI assisted, as I'm not familiar with that, so make sure you do a careful review.

File: runtime/src/built-ins.ts (hunk 12 above)
Comment (iquerejeta) [reply runtime/src/built-ins.ts]: Fair point - I've added the reduction

File: runtime/src/built-ins.ts (hunk 10 above)
Comment (iquerejeta) [reply runtime/src/built-ins.ts]: Great - then maybe merge this one first given that it is ready?

**Comment by iquerejeta** [comment]
@kmillikin @JosephDenman , ready for final review.

File: compiler/test.ss [hunk 14]
```diff
@@ -18086,6 +18086,28 @@ groups than for single tests.
               (safe-cast (tfield) (tunsigned 3) 3))))))
     )
 
+  (test ; jubjubSchnorrVerify with a simple message
```
JosephDenman, line 18089: Kevin at one point (before your corrections) mentioned that this program would fail:

```

ledger impure: Boolean;

export circuit test(): JubjubPoint {
  impure = true;
  return ecMulGenerator(jubjubScalarFromNative(0));
}
```

It's worth adding a regression test exactly of the test above.

JosephDenman (approved):
LGTM. Just please add the test before you merge.

PR #357, opened 2026-04-15
group dependabot PRs

pataei:
Groups dependabot PRs for each category based on semver.

Adds a workflow to assign reviewers to these PRs on a weekly rotation with a backup reviewer.


.github/dependabot.yml
```diff
@@ -5,15 +5,27 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    cooldown:
```
pataei, line 8: From Kent: "A cooldown period is essential even if there is a weekly update so that we (in theory) have time to learn of a malicious update before pulling it in."

kmillikin (approved):

kmillikin (approved):
Let's give it a try.

PR #396, opened 2026-04-22
Disable merge without changelog and compiler-version update

pataei:
**From a contributor's perspective:**
They open a PR that doesn't touch either the changelog nor the [compact-version.ss](http://compact-version.ss/). The check runs, it fails, the job exits 1, and the merge button is disabled. They have two paths: push a commit that edits changelog and [compact-version.ss](http://compact-version.ss/) (which makes the check to pass), or a maintainer adds the skip-changelog label (which makes the check to pass). Either way the red X becomes a green check and merge unlocks. This limits the ability to skip updating changelog to only folks with write access to our repo.

**Wiring into branch protection:**
The workflow by itself only reports a status. To actually block merging, we need SRE to Settings → Branches → Branch protection rules → add check-changelog.

### DIFF
--- a/.github/workflows/changelog-check.yml
+++ b/.github/workflows/changelog-check.yml
@@ -1,0 +1,63 @@
+name: Changelog Check
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+
+jobs:
+  check-changelog:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


kmillikin (approved):

pataei:
Asked SRE for adding `check-changelog` to branch protection rules: https://github.com/shieldedtech/shielded-sre/issues/167.

PR #400, opened 2026-04-22
Reserve event and log

pataei:
Please see changelog.


dybvig:
I do not think we should reserve `event`.  We can supply a predefined set of event structs as part of the standard library.

doc/compact-keywords.mdx
```diff
@@ -79,6 +79,7 @@ DO NOT EDIT: This file is automatically generated.
 - instanceof
 - interface
 - let
+- log
```
dybvig, line 83: Maybe this file needs to be rebuilt now that `event` is back in...

dybvig (approved):

PR #413, opened 2026-04-28
Add release notes for toolchain 0.31.0

kmillikin:


doc/release-notes/toolchain-0.31.0.md
```diff
@@ -1,111 +1,90 @@
...
-  ledger layout from the compiler output alone. Both exported and
-  non-exported fields are included since the full layout is required
-  to navigate the on-chain state tree and construct initial states.
+Compact supports iteration over a range of natural numbers, using the syntax `for (const i of start..end)`, where `i` is a variable name and `start` (inclusive) and `end` (exclusive) are expressions giving the iteration bounds.  The compiler must be able to determine the values of the iteration bounds at compile time.  Previously, the bounds were not allowed to be generic parameters.  Now this is allowed.
 
-## [Toolchain 0.30.105, language 0.22.101, runtime 0.15.101]
+### `fixup --line-length`
 
-### Added
+The `compact fixup` tool now supports a `--line-length` flag giving a column width for the fixed code.  Because fixes (e.g., renaming) can change the alignment of code, the `fixup` tool implicitly invokes the Compact formatter.  The formatter does support a `--line-length` flag, but it was not possible to tell the `fixup` tool the desired column width.
```
pataei: This change was applied to `fixup-compact`.  I'm not certain that the dev tool accepts all the flags accepted by a command, that is, `compact fixup --line-length` may not work even though `fixup-compact --line-length` does.

kmillikin: No, it does not.  We should change the devtools to be a thin wrapper around the underlying toolchain executable when a subcommand is provided by the toolchain.  If the toolchain executable doesn't have the command-line API we want, then we should change it directly in the toolchain.

I will just remove this section from the release notes.  We do not want to be telling developers to invoke toolchain executables directly, that leads to all kinds of version confusion and nonsense with PATH.

kmillikin:
I filed #414 to keep track of my suggested update to the devtools.

pataei (approved):

PR #421, opened 2026-05-04
Add typed 'outputs' to ZKIR v3

JosephDenman:
Just as the description says. The primary benefit of adding an `outputs` vector of types is that it greatly simplifies conformance checking in the CCCs feature. The conformance check must verify that the actual ZKIR of the callee circuit satisfies the circuit signature that the caller expects. Without typed outputs the conformance check would have to walk every `Output { val }` instruction in the callee, look up `val`'s type from prior instructions or inputs (which means a type-inference pass over the callee IR), and only then compare against the `ContractCall`s expected type descriptor. With typed `outputs`, this check becomes trivial.

Sibling PR in `midnight-ledger`: https://github.com/midnightntwrk/midnight-ledger/pull/480


kmillikin (approved):
I've merged `main` into it, pinned the `zkir-v3` dependencies to a ledger commit hash, and made some other small changes.

PR #442, opened 2026-05-15
feat(coip,events): Add 'Log Expression for Structured Event Emission' CoIP.

dzajkowski:
Adds a new CoIP document which lays out the introduction of a logging functionality for Compact language as part of the efforts to address [MPS-0005-events](https://github.com/midnightntwrk/midnight-improvement-proposals/mps/mps-0005-events.md).

### DIFF
--- a/coips/coip-xxxx.md
+++ b/coips/coip-xxxx.md
@@ -65,14 +65,13 @@ log(<struct instance>)
 - The expression returns the struct passed to it.
 - Multiple `log` calls per circuit are allowed.
 - Inside conditional branches, the log executes only if the branch is taken.
-- `pure` circuits must not contain `log`, it is a side effect.
 
 **Example:**
 
 ```compact
 circuit spend(...): [] {
     // ... spend logic ...
-    log(ShieldedSpend { nullifier: old_nullifier });
+    log(ShieldedSpend { nullifier: disclose(old_nullifier) });
 }
 ```
 
@@ -137,7 +136,7 @@ It compiles to a struct emission that downstream consumers can subscribe to.
 ```compact
 circuit spend(...): [] {
     // ... spend logic ...
-    log(ShieldedSpend { nullifier: old_nullifier });
+    log(ShieldedSpend { nullifier: disclose(old_nullifier) });
 }
 ```
 
@@ -166,7 +165,6 @@ This adds language surface for no semantic benefit — an event is a struct that
 ## Open Questions
 
 - What is the impact of `log` on the ZK circuit model? Adding `log` calls changes a circuit's public outputs. Does this have implications for proof generation or verification beyond additional instructions?
-- What is the hard cap on the number of events a contract can emit per circuit call?
 
 ## References
 
@@ -307,7 +305,7 @@ To detect spends, a consumer must poll the full state, diff Merklized structures
 circuit spend(dest_public_key: public_key, input_coin: coin_info): [] {
   // ... existing logic unchanged ...
 
-  log(ShieldedSpend { nullifier: old_nullifier });
+  log(ShieldedSpend { nullifier: disclose(old_nullifier) });
 }
 ```
 
--- a/coips/coip-xxxx.md
+++ b/coips/coip-xxxx.md
@@ -1,0 +1,316 @@
+---
+CoIP: X
+Title: Log Expression for Structured Event Emission
+Authors:
+  - Dominik Zajkowski (dzajkowski)
+Status: Draft
+Category: Language
+Created: 2026-05-15
+Requires: none
+Replaces: none
+---
+
+<!--
+This file is part of Compact.
+
+Copyright (C) 2026 Minokawa project contributors
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+## Abstract
+
+Compact has no mechanism for contracts to emit structured notifications about activity.
+
+This CoIP adds a `log` expression that accepts a struct instance.
+`log` is a disclosure site, the compiler enforces that all fields are already disclosed, same as ledger writes and circuit returns.
+The expression returns the struct passed to it.
+
+Phase 1 restricts `log` to the [standard event structs](#appendix-a-standard-events) shipped as part of the Compact standard library.
+
+## Motivation
+
+Compact contracts can modify ledger fields and return values from circuits, but have no way to signal what happened during execution.
+The only way to observe contract activity is to diff state between transactions.
+This requires knowledge of the contract's internal state structure, knowledge that is fragile and breaks when the contract evolves.
+Events still require knowledge of the contract's implementation, but provide an interface that can be kept stable more easily.
+
+Event emission is a well-established pattern in smart contract languages (Solidity, Ink!, Sway), proven useful over years of adoption.
+Compact's privacy-preserving design means a direct copy is not appropriate.
+This CoIP focuses on emission of already-disclosed data, a practical first step that improves the developer experience without introducing new disclosure concerns.
+
+## Specification
+
+### The `log` expression
+
+`log` is a new expression that accepts a struct instance.
+
+```compact
+log(<struct instance>)
+```
+
+**Semantics:**
+- `log` is a disclosure site — all field values must be already disclosed.
+- The expression returns the struct passed to it.
+- Multiple `log` calls per circuit are allowed.
+- Inside conditional branches, the log executes only if the branch is taken.
+- `pure` circuits must not contain `log`, it is a side effect.
+
+**Example:**
+
+```compact
+circuit spend(...): [] {
+    // ... spend logic ...
+    log(ShieldedSpend { nullifier: old_nullifier });
+}
+```
+
+### Phase 1 restrictions
+
+`log` accepts only structs from the [standard events](#appendix-a-standard-events) package shipped in the Compact standard library.
+The compiler rejects `log` calls with structs not in this set.
+A single log event must not exceed 1 KB serialized.
+These restrictions can be loosened in future CoIPs.
+
+### Standard events package
+
+The standard library ships a set of predefined event structs for common Midnight operations.
+See [Appendix A: Standard Events](#appendix-a-standard-events) for the full list of definitions.
+
+## Rationale
+
+**Why `log` instead of `emit`?**
+
+`emit` implies a runtime action with guaranteed delivery.
+`log` is more neutral — it records data, and what happens downstream is outside the language's scope.
+
+**Why disclosure enforcement?**
+
+This CoIP scopes `log` to already-disclosed values.
+Emitting private data is a different problem with different tradeoffs (encryption, topic-based filtering, recipient key management).
+Limiting to disclosed data keeps the first iteration simple and avoids new privacy concerns.
+
+**Why restrict to standard event structs in Phase 1?**
+
+A fixed set of known structs lets downstream tooling ship with built-in support.
+Unrestricted arbitrary structs require schema discovery mechanisms that are not yet in place.
+The `Misc` escape hatch covers custom use cases without blocking adoption.
+
+**Why return the struct?**
+
+Returning the struct promotes using standard event structs as data carriers — a value can be logged and used in the same expression.
+This encourages adoption of the standard events as a natural part of data flow rather than a separate bookkeeping step.
+
+## Backwards Compatibility
+
+`log` is a new keyword.
+If an existing contract uses `log` as an identifier, it will cause a compilation error.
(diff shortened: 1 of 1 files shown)


coips/coip-xxxx.md
```diff
@@ -0,0 +1,316 @@
...
+
+To detect spends, a consumer must poll the full state, diff Merklized structures, and trial-decrypt every ciphertext.
+
+## After
+
+```compact
+circuit spend(dest_public_key: public_key, input_coin: coin_info): [] {
+  // ... existing logic unchanged ...
+
+  log(ShieldedSpend { nullifier: old_nullifier });
```
pataei: This needs explicit disclosure: either `log(disclose (ShieldedSpend { nullifier: old_nullifier }))` or `ShieldedSpend { nullifier: disclose(old_nullifier) })`.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,316 @@
...
+**Arbitrary struct emission in Phase 1:** Requires schema discovery mechanisms (TypeScript descriptors, IR-based field resolution) that are not yet in place.
+Restricting to a known set enables built-in tooling support now; future CoIPs can loosen this.
+
+**Separate `event` type declaration:** Early drafts proposed a dedicated `event` keyword distinct from `struct`.
+This adds language surface for no semantic benefit — an event is a struct that gets logged.
+
+## Open Questions
+
+- What is the impact of `log` on the ZK circuit model? Adding `log` calls changes a circuit's public outputs. Does this have implications for proof generation or verification beyond additional instructions?
+- What is the hard cap on the number of events a contract can emit per circuit call?
```
pataei: Why does this matter?

coips/coip-xxxx.md
```diff
@@ -0,0 +1,316 @@
...
+```compact
+log(<struct instance>)
+```
+
+**Semantics:**
+- `log` is a disclosure site — all field values must be already disclosed.
+- The expression returns the struct passed to it.
+- Multiple `log` calls per circuit are allowed.
+- Inside conditional branches, the log executes only if the branch is taken.
+- `pure` circuits must not contain `log`, it is a side effect.
```
pataei: So `pure` circuits in Compact aren't defined as having general side effects. Impure circuits are those that might touch public state, call any witnesses, or call any other impure circuits. Pure circuits are those that are not impure and we presently assume that all native circuits are pure. 

With this definition, do you still want to reinforce `log` as impure? 

Some `log` expressions will be `impure` such as ones that are logging a ledger field or a witness or calling another impure circuit. But if a `log` expression is logging values such as a concrete `Bytes<32>` then it won't cause the circuit to be `impure`.

coips/coip-xxxx.md
dzajkowski: Thank you for catching that!

coips/coip-xxxx.md
dzajkowski: From the language perspective it does not. It should be captured in the MIP and enforced by the ledger.
I worry that it will make deploying Compact contracts more cryptic if at the point of writing there is no way to handhold the user, but I don't see practical ways to enforce this rule.

coips/coip-xxxx.md
dzajkowski: Fair. So the expectation should be something like:

> `pure` circuits must not contain `log` expressions that reference ledger fields or witnesses or call another impure circuit


?

pataei: I'd just drop the line re `pure` since `log` doesn't change the semantics of `pure` circuits.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,314 @@
...
+
+```compact
+struct ShieldedSpend {
+    nullifier: Bytes<32>          // hint: indexed
+}
+
+struct ShieldedReceive {
+    commitment: Bytes<32>,        // hint: indexed
+    contractAddress: Maybe<ContractAddress>,
+    ciphertext: Maybe<Bytes<512>> // hint: indexed
```
pataei: There are some discrepencies between this standard events and the one in https://github.com/shieldedtech/product/pull/93/files#diff-79e9c967f16962dbf27415c9336efca24f09ac41b0258bf6f86846db3646e4ef:
- `ciphertext` is indexed here but not there
- this puts most of the indexed fields before the unindexed ones (from a DB design this is preferable) but this isn't the case in the other proposal
- `UnshieldedSpend` has `domainSep` here but not there, same for `UnshieldedRecieve`
- there's a naming conventions difference, this uses `tokenType` but the other uses `token_type` (from Compact perspective `token_type` is fine since other struct declarations in Compact standard library follow this naming convention for fields, apologies I probably caused this confusion in my google doc).

dzajkowski: The design doc is superseded by the CoIP and the MIP. 

So as long as I did not fumble porting the data from the google doc we worked on, this should be the correct version.

> this puts most of the indexed fields before the unindexed ones (from a DB design this is preferable) but this isn't the case in the other proposal

I do not mind, happy to reorder to bring the indexed ones first.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,314 @@
...
+
+**Why restrict to standard event structs in Phase 1?**
+
+A fixed set of known structs lets downstream tooling ship with built-in support.
+Unrestricted arbitrary structs require schema discovery mechanisms that are not yet in place.
+The `Misc` escape hatch covers custom use cases without blocking adoption.
+
+**Why return the struct?**
+
+Returning the struct promotes using standard event structs as data carriers — a value can be logged and used in the same expression.
```
pataei, line 111: I disagree with this. Reasons:
- `log` is a side effect and from a language design point of view, it's unnatural for it to return the event struct. A user can simply write:
```
const x = e; // e is the event
log(x);
return x;
```
- From a consistent design point of view, `log` is very similar to a ledger write (including increment and decrement). Right now, in Compact, when you write `x = e` (or `x.write(e)` or `x += e`...) you don't expect to get `e` back, instead you expect (and this is what happens) to write `e` into ledger `x`. So we should stay consistent, either change ledger write operations to return the expression that they're writing to match `log` returning the event that is logging or change `log` to just be a side effect and return nothing.

The implementation I have right now is `log` has type `[]` (void in Compact in a way) and it doesn't return the event. Note from a workload perspective, changing this to `log` returning the event it's logging is not a big change (I'd just add a syntactic sugar similar to the code snippet above). However, changing the behavior of ledger writes will be a breaking change and requires more work.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,314 @@
...
+
+```compact
+log(<struct instance>)
+```
+
+**Semantics:**
+- `log` is a disclosure site — all field values must be already disclosed.
+- The expression returns the struct passed to it.
+- Multiple `log` calls per circuit are allowed.
+- Inside conditional branches, the log executes only if the branch is taken.
```
pataei, line 67: It's worth adding:
- `log` cannot be used in a constructor directly or indirectly.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,314 @@
...
+
+`log` is a new expression that accepts a struct instance.
+
+```compact
+log(<struct instance>)
+```
+
+**Semantics:**
+- `log` is a disclosure site — all field values must be already disclosed.
+- The expression returns the struct passed to it.
```
pataei, line 65: In a comment later I present my case of why this shouldn't be the case: https://github.com/LFDT-Minokawa/compact/pull/442/files#r3349114728

coips/coip-xxxx.md
```diff
@@ -0,0 +1,314 @@
...
+
+**Arbitrary struct emission in Phase 1:** Requires schema discovery mechanisms (TypeScript descriptors, IR-based field resolution) that are not yet in place.
+Restricting to a known set enables built-in tooling support now; future CoIPs can loosen this.
+
+**Separate `event` type declaration:** Early drafts proposed a dedicated `event` keyword distinct from `struct`.
+This adds language surface for no semantic benefit — an event is a struct that gets logged.
+
+## Open Questions
+
+- What is the impact of `log` on the ZK circuit model? Adding `log` calls changes a circuit's public outputs. Does this have implications for proof generation or verification beyond additional instructions?
```
pataei, line 167: We'll have a discussion on this in Compact community call, Wed Jun 3. 

My take: we still need a proof but it's a trivial proof. I cannot reason through having a proof that is actually meaningful and that might fail.

coips/coip-xxxx.md
pataei: Apologies for this one:

We need to add back that `log` makes a circuit impure. After a discussion with Kent, I learned that we have changed how we define pure. Currently we do define pure as having side effect. Even with the previous definition we still needed to consider `log` making a circuit impure since in the generated code we need to have access to context which pure circuits don't.

coips/coip-0003.md
```diff
@@ -0,0 +1,309 @@
...
+    // ... spend logic ...
+    emit(ShieldedSpend { nullifier: disclose(old_nullifier) });
+}
+```
+
+### Phase 1 restrictions
+
+`emit` accepts only structs from the [standard events](#appendix-a-standard-events) package shipped in the Compact standard library.
+The compiler rejects `emit` calls with structs not in this set.
+A single emit event must not exceed 1 KB serialized.
```
JosephDenman, line 83: Generally, I think its nice to explain reasoning whenever a constraint is imposed. Makes it easier to audit later if the subject is revisited. The 1 KB serialization limit should be explained.

coips/coip-0003.md
```diff
@@ -0,0 +1,309 @@
...
+`emit` is a disclosure site scoped to already-disclosed values.
+No new private data leakage is introduced.
+
+A contract can emit misleading events — `emit` records the contract author's claim, not independently verifiable facts.
+This is inherent to any event system.
+
+## How to Teach This
+
+For new users: `emit` works like `print` — it records a value for external observers without affecting contract state.
+Unlike `print`, it requires all values to be disclosed.
```
JosephDenman: We don't have a `print`, so using it to justify aspects of disclosure is confusing.

coips/coip-0003.md
```diff
@@ -0,0 +1,309 @@
...
+
+**Arbitrary struct emission in Phase 1:** Requires schema discovery mechanisms (TypeScript descriptors, IR-based field resolution) that are not yet in place.
+Restricting to a known set enables built-in tooling support now; future CoIPs can loosen this.
+
+**Separate `event` type declaration:** Early drafts proposed a dedicated `event` keyword distinct from `struct`.
+This adds language surface for no semantic benefit — an event is a struct that gets emitted.
+
+## Open Questions
+
+- What is the impact of `emit` on the ZK circuit model? Adding `emit` calls changes a circuit's public outputs. Does this have implications for proof generation or verification beyond additional instructions?
```
JosephDenman: I think this is a question that this document should and could answer concretely.

coips/coip-0003.md
```diff
@@ -0,0 +1,309 @@
...
+    sender: Either<ZswapCoinPublicKey, ContractAddress>,     // hint: indexed
+    tokenType: Bytes<32>,        // hint: indexed
+    amount: Uint<128>
+}
+```
+
+### Lifecycle Events
+
+```compact
+struct Paused {}
```
JosephDenman: Maybe a brief description of what these are? When they're emitted? The rest are self-explanatory. But, I don't know what `Paused` and `Unpaused` means.

coips/coip-0003.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Before
+
+```compact
+ledger nullifiers: Set<nullifier>;
+ledger commitments: HistoricMerkleTree<32, commitment>;
+ledger ciphertexts: Opaque<"Uint8Array">;
+
+circuit spend(dest_public_key: public_key, input_coin: coin_info): [] {
+  const source_secret_key = private$zk_secret_key();
```
JosephDenman: Maybe we should use camel case per the style guides?

coips/coip-0003.md
```diff
@@ -0,0 +1,309 @@
...
+  const fresh_commitment = commitment_from_coin_info(fresh_coin_info, dest_public_key.zk);
+  commitments.insert(fresh_commitment);
+  const ciphertext = disclose(context$encrypt(dest_public_key.encryption, fresh_coin_info));
+  ciphertexts.write(ciphertext);
+  private$remove_coin(input_coin);
+}
+```
+
+The values already public on-chain:
+- **`old_nullifier`** — disclosed hash, inserted into the public `nullifiers` Set.
```
JosephDenman:
```suggestion
- **`old_nullifier`** — disclosed hash, inserted into the public `nullifiers` set.
```

coips/coip-0003.md
```diff
@@ -0,0 +1,309 @@
...
+- **`fresh_commitment`** — disclosed hash, inserted into the public `commitments` MerkleTree.
```
JosephDenman:
```suggestion
- **`fresh_commitment`** — disclosed hash, inserted into the public `commitments` Merkle tree.
```

JosephDenman (changes requested):
The content is fairly straight forward. Mostly comments for legibility.

coips/coip-0003.md
dzajkowski: done.

coips/coip-0003.md
dzajkowski: done

coips/coip-0003.md
dzajkowski: good catch, thanks!

kmillikin (approved):

PR #447, opened 2026-05-19
Add kernel.caller() ledger operation

MicroProofs:
## Summary

Adds `kernel.caller()` to the ledger ADT surface. Returns the caller of a circuit
invocation as `Maybe<Either<ContractAddress, UserAddress>>`:

- `left(addr)` when called by contract `addr`
- `right(addr)` for a top-level call from user `addr`
- `None` when no caller can be determined (purely shielded top-level transaction,
  or unshielded inputs from multiple distinct owners)

## Implementation notes

The VM body of `caller()` needs the result type of the surrounding ADT op to size
its `concat` and produce the `None` default. The VM DSL has no `type-ref`, so
`expand-vm-code` now takes the result type as an explicit positional argument and
binds it to the implicit identifier `return_type` in the VM body. This mirrors how
`f` and `f-cached` were already bound.

Touched:

- `compiler/midnight-ledger.ss`: `caller` declaration.
- `compiler/vm.ss`: `expand-vm-code` gains `result-type`; `implicit-vm-bindings`
  centralizes the implicit identifier list used by `check-vm-code`.
- `compiler/ledger.ss`: `check-vm-code` call no longer carries `return_type`.
- `compiler/typescript-passes.ss`, `compiler/zkir-passes.ss`,
  `compiler/zkir-v3-passes.ss`: call sites pass the result type positionally.
- `compiler/test.ss`: inline fixture and JS tests for the three caller cases.
- `doc/ledger-adt.mdx`: documentation.
- `CHANGELOG.md`, `compiler/compiler-version.ss`, `compiler/language-version.ss`,
  `flake.nix`: version bump 0.31.102 -> 0.31.103.

## Test plan

- [x] `./compiler/go` passes (all 162 print-zkir, 167 print-zkir-v3,
      634 print-typescript tests across both `feature-zkir-v3 = #f` and `#t`)
- [x] JS fixture exercises `None`, contract caller, and user caller cases
- [x] `run-javascript` passes proof-data checks for both ZKIR v2 and v3


compiler/vm.ss
```diff
@@ -157,8 +157,13 @@
...
       (fluid-let ([query-src src])
         (expand-vm-expr e)))
-    (define (expand-vm-code src f f-cached arg-alist code)
-      (define expand-vm-expr (make-expand-vm-expr (cons* (cons 'f f) (cons 'f-cached f-cached) arg-alist)))
+    (define (expand-vm-code src f f-cached result-type arg-alist code)
+      (define expand-vm-expr
+        (make-expand-vm-expr
+          (cons* (cons 'f f)
+                 (cons 'f-cached f-cached)
+                 (cons 'return_type result-type)
```
dybvig: suggest: `return_type` => `result_type`

compiler/zkir-v3-passes.ss
```diff
@@ -872,7 +872,7 @@
               [env (list (cons 'emit-version event-version)
                          (cons 'emit-tag     event-tag)
                          (cons 'emit-payload (make-zkir-val payload-primitive-type* payload-alignment* triv*)))])
-         (assemble test '() '() '() src '() env vm-code instr*))]
+         (assemble test '() '() '() src '() #f env vm-code instr*))]
```
dybvig: I'm a bit concerned about the `#f` here for result-type. Maybe this should be:
```
(with-output-language (Lflattened Type) `(ty () ()))
```

compiler/midnight-ledger.ss
```diff
@@ -258,6 +264,49 @@
...
+     (swap [n 0])
+     ;; [context, effects, state, is_some=1, caller_slot]
+     (concat [cached #f] [n (rt-max-sizeof return_type)])
+     ;; [context, effects, state, (1, public_address)]
+     (jmp [skip 2])
+     ;; none-branch:
+     ;; [context, effects, state, caller_slot]
+     (pop)
+     ;; [context, effects, state]
+     (push [storage #f] [value (state-value 'cell (rt-null return_type))])
```
dybvig: Question: There's something subtle about the logic here, and I'm not following.  How does this produce 0, default?

dybvig (changes requested):
Just a couple of minor suggestions and a question.

compiler/midnight-ledger.ss
MicroProofs: rt-null creates the return type default value which produces the none = 0 case

compiler/midnight-ledger.ss
```diff
@@ -258,6 +264,49 @@
...
+     `left(addr)` when called by contract `addr`; `right(addr)` when this is the \
+     top-level call for user `addr` (uniquely identified by the transaction's \
+     unshielded inputs). None when no caller can be determined (e.g. a purely \
+     shielded top-level transaction, or one with unshielded inputs from \
+     multiple distinct owners). \
+     Maybe, Either, ContractAddress, and UserAddress are defined in CompactStandardLibrary."
+    ;; [context, effects, state]
+    ((dup [n 2])
+     ;; [context, effects, state, context]
+     (idx [cached #t] [pushPath #f] [path (list (align 6 1))])
```
kmillikin, line 285: I can't figure out where the 6 is coming from here.  I see 4 in the ledger code (https://github.com/midnightntwrk/midnight-ledger/blob/a1d1611c756c8973fd47b1e729686baf4dba6a35/onchain-runtime/src/context.rs#L309) and 5 in the spec (https://github.com/midnightntwrk/midnight-ledger/blob/a1d1611c756c8973fd47b1e729686baf4dba6a35/spec/contracts.md#context) which is probably out of date.

I think it needs a comment explaining where it comes from, because this code will break when the ledger call context format changes and we'll have to figure out how to update it.

compiler/midnight-ledger.ss
```diff
@@ -258,6 +264,49 @@
...
+     (eq)
+     ;; [context, effects, state, caller_slot, is_none]
+     (branch [skip 4])
+     ;; some-branch:
+     ;; [context, effects, state, caller_slot]
+     (push [storage #f] [value (state-value 'cell (align 1 1))])
+     ;; [context, effects, state, caller_slot, is_some=1]
+     (swap [n 0])
+     ;; [context, effects, state, is_some=1, caller_slot]
+     (concat [cached #f] [n (rt-max-sizeof result_type)])
```
kmillikin, line 304: Do we really need to parameterize this (and every other ledger operation) over `result_type` just for this?  We are already assuming that it is a representation of a specific Compact type so why can't we just hard-code the size and null value?

compiler/midnight-ledger.ss
```diff
@@ -81,6 +81,12 @@
             (type-ref UserAddress))
   "UnshieldedRecipient")
 
+(declare-ledger-type PublicAddress ()
```
kmillikin, line 84: Consider making this a standard library type (in Compact source) instead of just private to the ledger implementation.  It could also be a Compact alias.

compiler/midnight-ledger.ss
MicroProofs: Fair point, I talked with Kent and went down the route of adding result_type for future possible cases, but this seems to be a 1 off so hardcoding the null and size would also be fine.

dybvig: Yes.  The alternatives we considered resulted in much less elegant code, and this has potential future uses.

DavidM-D:
Just wanted to say how helpful this feature is. It makes our MPC deployment *significantly* simpler.

I'm sure it's generally true for other cross chain protocols.

JosephDenman:
The bytecode looks right to me. My concern is the `right` arm.

On chain, the ledger sets `caller = User(k)` for a root call whenever the intent's unshielded inputs all belong to `k`. Those inputs get added by the wallet during balancing, which in `midnight-js` happens after proving. On the client, nothing sets caller for the root: `createCircuitContext` has no option for it, `midnight-js`/`compact-js` never set it, and `WalletProvider` doesn't expose the unshielded key. So a root circuit that reads `kernel.caller()` in a call that takes NIGHT from the user (i.e. a deposit, the case you'd most want this for) records `None` in the transcript, the ledger computes `Some(User(k))`, and the tx fails with `ReadMismatch`. `left` is fine because the runtime sets the parent for callees and the contract rule wins on the ledger side. `None` is fine for shielded/dust-only calls.

On the ledger, the principle we'd need to prevent this kind of thing in the future is: "any `kernel` function needs to be computable from data present at contract execution time." Here, it depends on data only available after transaction balancing. To resolve this, we'd need an upstream `midnight-ledger` change. In the mean time, we have to decide whether we want to expose this feature knowing that it will fail in any root contract that manipulates unshielded tokens.

dybvig:
This doesn't sound like a security issue, so I would document the restriction (essentially, the behavior of `kernel.caller()` is undefined if used by a root contract.  It seems easy to avoid in user code with a Boolean argument to every circuit that uses it saying whether this is a cross-contract call or not (`false` when called directly by the Dapp and `true` when called from another contract).

JosephDenman (approved):

PR #466, opened 2026-05-28
CoIP: Contract interfaces, references, and calls

jonathan-sobel:
This Compact Improvement Proposal makes a major step forward in support for multi-contract systems by introducing contract interfaces, contract references (including storage in public ledger state), and cross-contract calls to Compact.


coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+<!--
+The abstract is a short (about 200 word) description of the issue being
+addressed and the proposed solution.
+-->
+
+In order to support the creation of multiple smart contracts that work
+together as a system, along with decentralized applications that use
+them, three major new features are added to Compact:
+1. contract interface types
+2. contract references in circuit parameters and ledger state fields
```
kmillikin: reference -> value ?  I sort of prefer "value" here.

Nothing prevents a contract reference (value?) from being returned from a circuit, does it?  I can think of ways that can be useful.  Likewise, nothing prevents them from being passed to or returned by witnesses (for circuits that can have witnesses).

coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+Clearly explain the problem and why the existing Compact language and tooling is
+inadequate to address the problem.
+-->
+
+Smart contracts are an integral part of the Midnight blockchain.
+Midnight provides the Compact language for defining contracts, and it
+uses zero-knowledge proofs to enable parts of a contract's execution
+to remain undisclosed.
+
+At the launch of Midnight's mainnet, it was already possible to write
```
kmillikin: I don't love using "mainnet" as a word (but maybe it is, after all).  The official name of Midnight's mainnet is "Midnight Mainnet" though so I think I'd rather see "launch of Midnight Mainnet".

coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+The Compact grammar already includes a production for "external
+contract declaration."  This is replaced by a production for
+"interface declaration."  The only syntactic change is the use of the
+keyword `interface` in place of `contract`.  For clarity in the
+grammar, the production's nonterminal should be renamed from
+_contract-declaration_ to _interface-declaration_, and the identifier
+for the interface should be renamed from _contract-name_ to
+_interface-name_.  Also, _interface-name_ should be added as an alias
```
kmillikin: Added as an alias?

coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+
+Thus, the production _interface-declaration_ looks like
+
+> `export`^opt `interface` _interface-name_ `{` _circuit-declaration_ `;` ... `;` _circuit-declaration_  `;`^opt `}` `;`^opt
+
+and likewise for the comma-separated version.  Note that
+_circuit-declaration_ is defined to require a "simple" parameter
+list.  The circuits in contract interfaces must not have generic
+parameters.
```
kmillikin: This production (simple-parameter-list) also disallows patterns.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+
+Any contract whose set of exported circuits is a superset of the
```
kmillikin: As an implementer, I would want more detail here so this was not left up to my interpretation.

A contract has a set of declared exported circuits, which generates a set of signatures (by erasing the parameter names and replacing patterns with simple identifiers).  An interface also generates a set of signatures (by erasing the parameter names).  Then we need all the corresponding signatures to be equal (or can the contract's signature be a subtype of the interfaces?).

coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+One may imagine a future version of Compact having quite rich support
+for multi-contract systems, with dynamic discovery of the actual
+circuit code that implements an interface and with support for witness
+functions and private state in called circuits.  This proposal
+describes only the first step in that direction.
+
+Specifically, two major limitations are incorporated into this first
+proposal for multi-contract system support:
+
+1. An interface named `T` can be implemented only by a contract
```
kmillikin: The specification does not say this ("interface implementation is structural").  The true situation is subtle: we will use the JS and ZKIR implementation from `T.compact` (and the prover keys) for every circuit invoked on contract value of interface type `T`.

We don't necessarily need all that detail here, just to not say something that's not (quite) right.  A later section can talk about how contract calls are resolved to an implementation.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+   defined in the file `T.compact`.  That is, the code for a contract
+   that implements interface `T` may be dynamically loaded, but it
+   will be found by assuming that the code for concrete contract `T`
+   is already made available by the DApp.
+2. Contracts to be called by other contracts must not define witness
```
kmillikin: This isn't statically checked as far as I can see.  We need to ensure that if a contract value (just an address) is passed to a circuit and a circuit is invoked, that that call will fail at runtime if it tries to invoke a witness (say, by passing the JS implementation something that will fail in place of the witnesses it would normally get).

coips/coip-xxxx.md
```diff
@@ -0,0 +1,351 @@
...
+On the other hand, when a contract is intended to implement some
+interface, it is useful to assert that expectation.  This is
+accomplished with a new program element to assert that the current
+contract implements a specific interface.  Here is an example.
+
+```
+contract implements Adder;
+```
+
+This declaration applies to the contract being defined in the current
```
kmillikin: So this is a top-level only program element, it can't appear in a module?  Or can I import this from a module?

I suppose that `Adder` needs to be an interface in scope, but we should say that.

kmillikin (commented):
I'm trying to read this as a third party implementer of the feature.

I think most of my comments are actually addressed later in the proposal, but we should try to structure the early part to clarify that, even if only by letting me know that there is a simplification that will be clarified in [forward reference].

coips/coip-xxxx.md
jonathan-sobel: The idea is that contracts can *refer* to other contracts, thus "references". Talking about "contract values" might imply copy-by-value or something like that to some readers, and "pointer" implies memory addresses to many people. "References" seemed like a safe way to talk about pointers without saying pointers.

On the other hand, if it's more in keeping with other Compact docs, I can also change the language throughout the CoIP, including the title ("Contract Interfaces, Values, and Calls"), to value terminology.

As for mentioning parameters and fields, those were meant to be examples, but I agree that it sounds like limitations. I'll fix that.

coips/coip-xxxx.md
jonathan-sobel: Rewriting to eliminate the problem: "Prior to the improvements described in this proposal..."

coips/coip-xxxx.md
jonathan-sobel: Odd text editing mistake. Fixing it and referring to _interface-name_ as a "meta-variable" instead of an "alias."

coips/coip-xxxx.md
jonathan-sobel: This was meant as the semantic explanation for that syntactic restriction. Rewriting to make that clearer:
> Note that _circuit-declaration_ is defined to require a "simple" parameter list, because the circuits in contract interfaces are not allowed to have generic parameters.

coips/coip-xxxx.md
jonathan-sobel: You're right. I'm adding a great deal more precision in this section, defining signatures carefully and giving examples of subtyping.

coips/coip-xxxx.md
jonathan-sobel: Yes, it's strangely difficult to say something correct about this in this introductory section without giving the full specification. I'm rewriting this item, hoping to do better.

coips/coip-xxxx.md
jonathan-sobel: It seems like it *could* be statically checked, with a simple requirement that when a contract has witnesss functions, it fails (by rule) to implement any interfaces, so type checking would fail if you violate the constraint, at least within Compact. But since contract values will originate from outside Compact, the dynamic approach you describe will also be necessary.

Anyway, at this point in the write-up, the restriction is directed more at the programmer: You can't have witness functions in a contract that you want to call from another contract.

coips/coip-xxxx.md
jonathan-sobel: I *think* an `implements` declaration could appear in a module. That's why I wrote the specifiction this way. Whatever contract you are in the middle of defining when you hit the declaration, even if you are inside a module, is the one it applies to.

And yes, I'll add something about `Adder` having to be in scope.

coips/coip-xxxx.md
```diff
@@ -0,0 +1,588 @@
...
+implemented.  The intended behavior should be clearly described and unambiguous.
+-->
+
+### Contract Interfaces
+
+A new form is added to Compact's program elements: the contract
+interface declaration.  Here is an example.
+
+```compact
+interface Adder {
```
dybvig: This conflicts with the TypeScript use of `interface`, might be confusing for TypeScript users, and might preempt future extension in the direction of adding TypeScript interfaces to Compact.  I wouldn't have a problem sticking with `contract`, but if others do, maybe we should use `contract interface` instead.

coips/coip-0002.md
```diff
@@ -0,0 +1,588 @@
...
+of its parameters, plus its return type.  Changing the parameter names
+or using destructuring patterns in a circuit definition does not
+change its signature, because the signature is defined only by the
+parameter and return *types*.  The circuit declarations in an
+interface also have signatures defined by their types.  For example,
+the signature of `addTo` in the `Adder` interface above is
+
+> `Uint<64>` &rarr; `Uint<64>`
+
+One circuit signature *S* is a **subtype** of another *T* when
```
dybvig, line 165: Subtyping for interface circuits is problematic.  We don't presently guarantee the property that type S ⊆T ⇒ S has the same representation as T, and there's no where to put the conversion code for cross-contract calls.  It can't go in the calling contract, because it doesn't know that the callee has different types, and it can't go in the called contract, because the same circuit might be used to satisfy more than one interface and/or be called from TypeScript.  So unless we're willing to commit now to the same representation property, we should require the implementation of an interface circuit to have the same parameter and return types.

coips/coip-0002.md
```diff
@@ -0,0 +1,588 @@
...
+ways to find the code were suggested.  Here are a few.
+
+- The application provides a map from contract addresses to filesystem
+  paths, saying where to find the code for a given deployed contract.
+- The application provides a map from interface names to filesystem
+  paths, saying where to find the code that implements a given
+  interface.
+- The application provides a URL for an external service that supplies
+  one of the preceding mappings.
+- Some representation of circuit code is stored in the ledger when a
```
dybvig, line 565: Just to keep in mind for the future, this representation either cannot be the just binary zkir that is currently stored in the ledger (as of ledger 9) or we cannot support cross-contract calls to circuits that don't require proofs, because there is no binary zkir stored in the ledger for such circuits.

coips/coip-0002.md
jonathan-sobel: Oh, wow. I had no idea that using a value with a subtype of the expected value in Compact might have additional overhead from implicit conversions.

That forces the use of subtyping relationships in the language to be determined partly by the ability of compiler to inject conversion operations where needed. I'd like to propose that this be fixed in a future version of the language. It feels strange to say B is a subtype of A, but I cannot use B in some places where A is expected.

If that's the current situation, I'll remove all discussion of subtypes from the CoIP and require an exact signature match in order to implement an interface.

coips/coip-0002.md
jonathan-sobel: It will be necessary to revisit this in order to move forward into the next stages of multi-contract systems. I'll edit the text a little so it commits less strongly to that precise solution.

coips/coip-0002.md
kmillikin: I think we now (soon) do have the property that subtyping implies same representation, but I can easily see us breaking it again.  (For example: struct subtyping doesn't have that property any more, because at least in circuit struct types are "fully unboxed" so their representation is a flat sequence of ZKIR-typed values and we'd probably allow fields to be treated as if unordered.)

But I don't think the situation is as dire as it seems.  What we need is the weaker property that "S is a subtype of T" implies that there is a conversion (cast) from S to T that the runtime is able to perform.  The compiler does not insert this conversion, the runtime that makes the cross contract call does.  At the point of the call, it needs to know the signature of the called circuit and the types of the arguments/expected return value.

coips/coip-xxxx.md
kmillikin: After some private discussion, I'm happy with `contract` as well.

coips/coip-0002.md
jonathan-sobel: The question is really: Are we willing to make a long-term commitment to allowing programmers to rely on subsumption, with the compiler and/or runtime ensuring that any necessary conversions occur?

But a question of that magnitude should not be answered in this improvement proposal. I'll leave subtyping out of this one, and a different proposal can argue for general (or specific) subtype subsumption.

kmillikin (approved):

PR #476, opened 2026-06-05
feat(coip,language): Add 'Explicit visibility modifiers for ledger fields and circuits' CoIP

0xisk:
## Summary

CoIP proposing explicit visibility modifiers (`private` / `internal` / `public`) for ledger fields and circuits, replacing the overloaded `export` keyword.

Today `export` controls two unrelated things: whether a binding is importable by other modules, and whether it appears in the contract's generated TypeScript API. A module's `export` makes a binding importable but does not surface it in the generated TypeScript, so contracts composed from several modules must re-export each item by hand (`export { M_state };`). That list is verbose and silently wrong when an item is forgotten.

The proposal:

- `private` — module-scoped only (today's default).
- `internal` — importable by other modules, not in the TypeScript API.
- `public` — importable and emitted to the TypeScript API, transitively, with no manual re-export.

Emission stays opt-in: only what the module author marks `public` is surfaced, so nothing is exposed by accident.

## Background

Originates from discussion [OpenZeppelin/midnight-apps#272](https://github.com/OpenZeppelin/midnight-apps/discussions/272) and the review thread it grew out of.

### DIFF
--- a/coips/coip-xxxx.md
+++ b/coips/coip-xxxx.md
@@ -1,0 +1,296 @@
+---
+CoIP: X
+Title: Explicit visibility modifiers for ledger fields and circuits
+Authors:
+  - Iskander Andrews (0xisk), OpenZeppelin
+Status: Draft
+Category: Language
+Created: 2026-06-05
+Requires: none
+Replaces: none
+---
+
+<!--
+ This file is part of Compact.
+ Copyright (C) 2026 Minokawa project contributors
+ SPDX-License-Identifier: Apache-2.0
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


dybvig:
Thank you for your proposal.

We discussed this in the community call today.  The consensus is that imported modules should not be able to bypass the existing import/export mechanism and force a top-level export.  That is, the contract should remain in control over the set of top-level exports. Allowing modules to directly effect top-level exports also prevents the importing contract from dealing with name clashes in the exports of two or more modules.

We suggest an alternative solution, which is that the `export` program element be extended to allow the export of all of a module's bindings.  This might look like:
```
export { import M }
```
That is, an `export` form can contain within it an `import` form, and the effect is to export all of the identifiers that would be imported by the `import` form.

With this syntax, top-level export of the **n** exports of each of **m** modules requires only **m** rather than **m** x **n**`export` subforms, and it does not require the set of exports to be updated in multiple places.

I'm inclined to say that when this syntax is used at the top level of a contract, exports of the imported module that cannot be exported from the top level (e.g., circuits with type parameters) are silently ignored for convenience.  (Explicit export of such a thing would still be a static error.)

Separately, we have in mind extending the `import` form to allow import of all but a specified set of a module's exports and to allow individual exports to be renamed.  These features would also be supported by `import` forms embedded within `export` forms.

PR #485, opened 2026-06-10
Front-end updates for cross-contract calls

dybvig:
This PR builds on JosephDenman/ccc:
- eliminates static checks for compatibility of contract types (and no looking for or reading compact-info.json files)
  - NB: comments out some tests-e2e tests that assume these static checks are done
- adds a new `contract implements type` program element that verifies that the current contract implements the given contract type
- injects subcontract imports into the generated JavaScript code based on the contracts actually called
- contract-info.json now includes full types for called contracts rather than just contract names


compiler/test.ss
```diff
@@ -24912,6 +24936,220 @@ groups than for single tests.
...
+      )
+    (oops
+      message: "~a:\n  ~?"
+      irritants: '("testfile.compact line 1 char 1" "contract implements failure:\n  this contract exports a circuit named ~s, but\n  its return type is ~a rather than ~a" (foo "[]" "Uint<16>")))
+    )
+
+  (test-group
+    ((create-file "C1interface.compact"
+       '(
+         "module C1interface {"
```
JosephDenman, line 25135: I still see test assertions like:

```
("testfile.compact line 1 char 1" "cannot export ~a (~s) from the top level" ("contract type" C))
```

Why is it an error to export a contract type from the type level but not, e.g., a `struct`?

JosephDenman, line 25135: While I was playing around with this, I realized I don't know why our current `import` statements seem to require that, to import a module, the name of the file defining the module must correspond to the name of the module. 

For example:

```
// A.compact
module def {
  export contract C {
    circuit foo(x: Bytes<32>): [];
    pure circuit barr(): Bytes<32>;
  }
}
```

and

```
// B.compact
import { C } from def;
contract implements C;
```

shows

```
Exception: B.compact line 1 char 1:
  failed to locate file "def.compact"
```

Is there a reason to not all the name of the file and the module it defines to vary freely? Kind of counter-intuitively, I would also expect to get a corresponding error if `A.compact` was renamed to `Def.compact` (with the first letter capitalized), but then I the error:

```
 Exception: def.compact line 3 char 3:
  contract implements failure:
  this contract does not export a circuit named foo
```

I would expect an error about the `implements` failure because the circuits don't match. Not only does the case-mismatch not trigger an error, as I would expect would be the same class of error as when the file was named `A.compact`. But, also `def.compact` isn't an actual file we tried to compile. And, I would expect the error to point to the fact that `B.compact` was in fact declaring something incorrect - that `B.compact` made an untrue `contract implements C` declaration. Maybe the second issue is just because the `import` in `B.compact` does some in-lining. Even then, the error isn't easy to decode. And, the `Def.compact`/`def.compact` asymmetry seems obviously wrong.

JosephDenman (commented):
Working on understanding these changes still, but I'll add the comments I have so far and continue tomorrow.

compiler/test.ss
dybvig: Exporting something from a program's top level causes it to be represented by a corresponding entry in the generated TypeScript/JavaScript code.  For example, exporting a struct causes a corresponding type to appear in the .d.ts file, and exporting a circuit causes a corresponding function to appear in the Circuits object.  If there is some appropriate counterpart for a contract type declaration that we should be putting in the .d.ts file, we should allow them to be exported from the program top level; otherwise, it's better to reject top-level exports of contract types.

dybvig: > Is there a reason to not all the name of the file and the module it defines to vary freely?

Well, first, it would be expensive.  The compiler would have to open every `.compact` file in the search path and parse it at least minimally until it finds the module, without complaining about syntax errors. That could be expensive.  Also, a typo could prevent the compiler from finding the intended file, and either the wrong module would be found or the developer would get a "module not found" error instead of a useful error about the module definition being incorrect in some way.

> I would also expect to get a corresponding error if A.compact was renamed to Def.compact (with the first letter capitalized)

You would on a case-sensitive filesystem, but on case-insensitive filesystems, `Def.compact` and `def.compact` are effectively the same.

Regarding the `contract implements failure`, I originally thought about allowing the `contract implements` form to appear only at the top level of a contract, but I decided that it might be useful to package a `contract implements` form with a contract-type declaration in the same module so that importing the module prevents the importing contract from compiling unless it fully implements the contract.  I agree that the message is misleading and the even the concept might be confusing, so either we should work on the message or go with my original thinking.

PR #491, opened 2026-06-15
Add basic support for Secp256k1Point

kmillikin:
They are supported in ZKIR v3 only, through to the ZKIR v3 backend.  The operations `ecAdd`, `ecMul`, and `ecMulGenerator` are implemented and generate correct ZKIR v3 output.

It's not yet implemented to extract the components of a point, nor can they appear in the ledger or be returned from witnesses.

### DIFF
--- a/runtime/src/compact-types.ts
+++ b/runtime/src/compact-types.ts
@@ -275,7 +275,7 @@ export const CompactTypeSecp256k1Base: CompactType<bigint> = {
   },
 
   toValue(value: bigint): ocrt.Value {
-    if (value > MAX_SECP256K1_BASE) {
+    if (value < 0n || value > MAX_SECP256K1_BASE) {
       throw new CompactError('expected Secp256k1Base');
     }
     let res: ocrt.Value = [];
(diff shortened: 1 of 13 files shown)


kmillikin:
Adding @JosephDenman as a reviewer to take a look at the JS descriptors for the base and scalar fields.

runtime/src/compact-types.ts
```diff
@@ -242,6 +243,96 @@ export const CompactTypeField: CompactType<bigint> = {
...
+      res = (res << 64n) | ocrt.valueToBigInt([value[i]]);
+    }
+    if (res > MAX_SECP256K1_BASE) {
+      throw new CompactError('expected Secp256k1Base');
+    }
+    return res;
+  },
+
+  toValue(value: bigint): ocrt.Value {
+    if (value > MAX_SECP256K1_BASE) {
```
kmillikin: I guess this should also ensure `value` is non-negative.

dybvig (approved):

PR #497, opened 2026-06-17
ci: on-demand dev publish of compactc + compact-runtime

skylarsimoncelli:
# ci: on-demand dev publish of compactc + compact-runtime

## What

Adds a manually-triggered (`workflow_dispatch`) workflow that publishes an
installable Compact toolchain for **any branch/commit**, without cutting a
tagged release:

- **compact-runtime** -> `@midnight-ntwrk/compact-runtime@<base>-dev.<short-sha>`
  on GitHub Packages, under npm dist-tag `dev` (never moves `latest`).
- **compactc binary** -> GitHub **prerelease** `compactc-dev-<short-sha>` on this
  repo, per-architecture zip assets.

Plus `doc/dev-builds.md` documenting the trigger contract, coordinates, consume
instructions, and retention policy.

## Why

compactc / compact-runtime currently publish no installable artifact off-tag, so
every downstream consumer (midnight-js, midnight-node, compact-js,
midnight-toolkit) clones this repo and Nix-builds compactc on every CI run
(~15 min each, paid N times; pulls a Rust/Nix toolchain into pure-TS repos;
breaks silently on toolchain drift). midnight-js#978 and midnight-node already
carry this workaround independently. This breaks the "merge to main -> cut a
release" dependency for downstream integration (e.g. Events e2e validation).

Ref: shieldedtech/shielded-sre#324.

## How it works (reuses existing machinery, no forked build logic)

- Binary: fans out to the existing reusable `release-build.yml` /
  `release-test.yml` per platform (Linux x86_64 always built + the publish gate;
  Linux arm / macOS arm / macOS intel optional via inputs to keep on-demand runs
  cheap), then publishes a prerelease via `softprops/action-gh-release` with
  `secrets.MIDNIGHTCI_REPO` -- same pattern as `internal-release.yml`.
- Runtime: `nix build .#runtime.forPublish` (the existing publishable tarball),
  repacked with the dev version string, `npm publish --tag dev` to
  `https://npm.pkg.github.com/` with `secrets.MIDNIGHTCI_PACKAGES_WRITE`. The
  version is applied at publish time (not committed) per the flake.nix note about
  prerelease runtime versions.

## Reviewer notes / things to confirm

- **`MIDNIGHTCI_PACKAGES_WRITE` must have `write:packages` on the `midnight-ntwrk`
  org.** It is currently used only for npm reads + GitHub Release writes; the
  runtime `npm publish` is the one net-new capability. First in-repo
  `workflow_dispatch` run confirms it.
- No push/PR/schedule trigger -- manual only, runnable by anyone with write
  access.
- Automated pruning of old `dev.*` versions is intentionally left as a
  fast-follow; retention intent is documented in `doc/dev-builds.md`.

## Acceptance criteria (from shielded-sre#324)

- [x] Installable artifact producible on demand, not only on tagged releases.
- [x] Stable coordinate to pin to (`@midnight-ntwrk/compact-runtime@x-dev.<sha>`,
      `compactc-dev-<sha>`).
- [x] Retention policy documented (`doc/dev-builds.md`).
- [x] Documented trigger contract -- who can run it, inputs, where artifacts land.

### DIFF
--- a/.github/workflows/dev-publish.yml
+++ b/.github/workflows/dev-publish.yml
@@ -12,10 +12,11 @@
 #   - Input `branch` selects the ref to build and publish from.
 #
 # Coordinates produced (stable, pinnable)
-#   - npm:    @midnight-ntwrk/compact-runtime@<base>-dev.<short-sha>  (dist-tag `dev`)
+#   - npm:    @midnight-ntwrk/compact-runtime@<base>-dev.<sha>  (dist-tag `dev`)
 #             registry https://npm.pkg.github.com/  (scope @midnight-ntwrk)
-#   - binary: GitHub *prerelease* on this repo, tag `compactc-dev-<short-sha>`,
-#             assets compactc_dev-<short-sha>_<arch>.zip
+#   - binary: GitHub *prerelease* on this repo, tag `compactc-dev-<sha>`,
+#             assets compactc_dev-<sha>_<arch>.zip
+#   <sha> is the full 40-char commit SHA.
 #
 # Retention
 #   - npm `dev.*` prereleases publish under dist-tag `dev` and never move `latest`.
@@ -69,7 +70,7 @@ jobs:
       contents: read
     outputs:
       branch: ${{ inputs.branch }}
-      short_sha: ${{ steps.coord.outputs.short_sha }}
+      sha: ${{ steps.coord.outputs.sha }}
       bin_tag: ${{ steps.coord.outputs.bin_tag }}
       runtime_version: ${{ steps.coord.outputs.runtime_version }}
       linux_arm: aarch64-unknown-linux-musl
@@ -86,13 +87,15 @@ jobs:
         id: coord
         run: |
           set -euo pipefail
-          SHORT_SHA="$(git rev-parse --short HEAD)"
+          # Use the full 40-char commit SHA so the coordinate is unambiguous and
+          # cannot collide (short SHAs can, in a repo this size).
+          SHA="$(git rev-parse HEAD)"
           BASE_VERSION="$(node -p "require('./runtime/package.json').version")"
-          RUNTIME_VERSION="${BASE_VERSION}-dev.${SHORT_SHA}"
-          echo "short_sha=${SHORT_SHA}" >> "$GITHUB_OUTPUT"
-          echo "bin_tag=dev-${SHORT_SHA}" >> "$GITHUB_OUTPUT"
+          RUNTIME_VERSION="${BASE_VERSION}-dev.${SHA}"
+          echo "sha=${SHA}" >> "$GITHUB_OUTPUT"
+          echo "bin_tag=dev-${SHA}" >> "$GITHUB_OUTPUT"
           echo "runtime_version=${RUNTIME_VERSION}" >> "$GITHUB_OUTPUT"
-          echo "compactc binary tag: dev-${SHORT_SHA}"
+          echo "compactc binary tag: dev-${SHA}"
           echo "runtime npm version:  ${RUNTIME_VERSION}"
 
   build-linux-x86:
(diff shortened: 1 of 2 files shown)


.github/workflows/dev-publish.yml
```diff
@@ -0,0 +1,277 @@
...
+            - Runtime npm: `@midnight-ntwrk/compact-runtime@${{ needs.setup.outputs.runtime_version }}` (dist-tag `dev`)
+
+            Not a supported release. See doc/dev-builds.md for the retention policy.
+
+  publish-runtime:
+    name: Publish compact-runtime dev npm package
+    if: ${{ inputs.publish_runtime }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
pataei, line 244: Shouldn't this be a `write` permission?

.github/workflows/dev-publish.yml
```diff
@@ -0,0 +1,277 @@
...
+          pattern: dev-*
+          merge-multiple: true
+
+      - name: Create dev prerelease on LFDT-Minokawa/compact
+        id: create_release
+        uses: softprops/action-gh-release@b4309332981a82ec1c5618f44dd2e27cc8bfbfda # v3.0.0
+        with:
+          repository: LFDT-Minokawa/compact
+          tag_name: compactc-${{ needs.setup.outputs.bin_tag }}
+          target_commitish: ${{ needs.setup.outputs.short_sha }}
```
pataei: I'm not an expert here but isn't it safer to rely on the full hash instead of short hash?

pataei (dismissed):
Re:

> MIDNIGHTCI_PACKAGES_WRITE must have write:packages on the midnight-ntwrk
org. It is currently used only for npm reads + GitHub Release writes; the
runtime npm publish is the one net-new capability. First in-repo
workflow_dispatch run confirms it.

I don't have access to the permissions of `MIDNIGHTCI_PACKAGES_WRITE`. So once the other issues are addressed, I'm happy to merge this to run the workflow and see what happens

.github/workflows/dev-publish.yml
skylar-simoncelli: Good question, but `contents: read` is correct here. `npm publish` authenticates with the `MIDNIGHTCI_PACKAGES_WRITE` **PAT** (passed as `NODE_AUTH_TOKEN`), not `GITHUB_TOKEN` — and the job-level `permissions:` block only governs `GITHUB_TOKEN`. On top of that, the package scope is `@midnight-ntwrk`, a *different org* than this repo (`LFDT-Minokawa`), so `GITHUB_TOKEN`'s `packages: write` couldn't publish there even if granted. Same reasoning for the binary prerelease (uses the `MIDNIGHTCI_REPO` PAT). I've added comments on both publish jobs spelling this out.

.github/workflows/dev-publish.yml
skylar-simoncelli: Agreed — switched to the full 40-char commit SHA for the npm version, the prerelease tag, and `target_commitish`. Short SHAs can collide in a repo this size, and an ambiguous coordinate defeats the "which compactc was this tested against?" goal. Fixed in fd441cf.

skylar-simoncelli:
Thanks for the review @pataei. Pushed fd441cf addressing both:

1. **Full commit SHA** — the dev coordinate (npm version, prerelease tag, `target_commitish`) now uses the full 40-char SHA instead of the abbreviated one, so it's unambiguous and collision-free.
2. **Publish permissions** — kept `contents: read` (correct: both publish paths authenticate via PATs, not `GITHUB_TOKEN`, and target the `@midnight-ntwrk` org which `GITHUB_TOKEN` can't write to anyway) and added inline comments on both publish jobs explaining why.

Docs updated to match. Re the `MIDNIGHTCI_PACKAGES_WRITE` scope question — that's the one thing only an in-repo `workflow_dispatch` run can confirm, so whenever you're ready to merge-and-test, that run will tell us if the token has `write:packages` on `midnight-ntwrk`.

.github/workflows/dev-publish.yml
ryjones: I suggest setting up OIDC publishing for NPM. If you need help, reach out.

kmillikin: Thanks for the suggestion @ryjones.  I'll create an issue to track it.

kmillikin (approved):

.github/workflows/dev-publish.yml
kmillikin: This is tracked in #509.

PR #500, opened 2026-06-17
Add JS encoding Secp256k1Point, fix the field encodings

kmillikin:
The field encodings need to have 1 subtracted before encoding.


kmillikin:
This was incorporated into #519

PR #520, opened 2026-06-23
refactor: add secp related compact types and fix ecdsa.test.ts

golddydev:


compiler/midnight-natives.ss
```diff
@@ -59,22 +59,12 @@
...
   "__compactRuntime.secp256k1ScalarMul"
   ([x Secp256k1Scalar (discloses "a product including")]
    [y Secp256k1Scalar (discloses "a product including")])
   Secp256k1Scalar)
 
-(declare-native-entry circuit inv
-  "__compactRuntime.secp256k1ScalarInv"
-  ([s Secp256k1Scalar (discloses "the inverse of")])
-  Secp256k1Scalar)
-
```
iquerejeta: I can't find these duplicates - I see the implementations for the base field, but not the scalar field. Even if we don't need/use them for ECDSA, they should be added IMO

examples/ecdsa/example_one.compact
```diff
@@ -58,3 +58,13 @@ export circuit proveBitcoinSignature(
...
+
+// Multiply two secp256k1 scalars.
+// The native `mul` is the special case that reduces the product modulo the
+// secp256k1 group order n (SECP256K1_SCALAR_MODULUS), which is the correct
+// modulus for the scalar field. The bare `*` operator would instead lower to
+// the runtime's `mulField`, reducing modulo the proof system's BLS12-381 field
+// modulus -- the wrong modulus for a Secp256k1Scalar.
+export circuit scalarMul(x: Secp256k1Scalar, y: Secp256k1Scalar): Secp256k1Scalar {
+  return mul(x, y);
+}
```
iquerejeta: Why add this wrapper function?

runtime/src/built-ins.ts
```diff
@@ -463,7 +463,8 @@ export function alignedConcat(...values: ocrt.AlignedValue[]): ocrt.AlignedValue
  * The returned value is in the range [0, JUBJUB_SCALAR_MODULUS).
  */
 export function jubjubSampleScalar(): bigint {
-  return ocrt.valueToBigInt(ocrt.jubjubSampleScalar());
+  return 0n;
+  // return ocrt.valueToBigInt(ocrt.jubjubSampleScalar());
```
iquerejeta: Leftover?

iquerejeta (commented):
Looks good to me! 👍

examples/ecdsa/example_one.compact
golddydev: Oh, this is only for testing mul for scalar, checking modulus is correct.
I will remove this, other tests can be sufficient enough.

compiler/midnight-natives.ss
golddydev: They are under `zkir-v3-natives.ss`
https://github.com/golddydev/compact/blob/gd/test-secp/compiler/zkir-v3-natives.ss

runtime/src/built-ins.ts
golddydev: Fixed.

PR #522, opened 2026-06-23
Update coip-0002 to talk about types, rather than interfaces

jonathan-sobel:
The original version of coip-0002 used "interface" and "implements" terminology. This version uses terminology more consistent with the rest of Compact: the `contract` form defines a new contract type (parallel to `enum` and `struct`), and contract values have contract types.

Also, the original version completely prohibited the definition of witness functions in called contracts. This new version weakens the restriction, so that witness functions can be defined in any of the contracts, just not called by circuits that are called from other contracts (more precisely, circuits that call witnesses cannot be used to satisfy a contract type).


coips/coip-0002.md
```diff
@@ -69,37 +69,37 @@ contract calling the circuits of other contracts.
...
-1. When a contract holds a reference to another contract of interface
+1. When a contract holds a reference to another contract of
    type `T` and calls one of its circuits, the implementation of the
    circuit is found by assuming that it was produced by compiling
    `T.compact` and that it is available in the same application
    context as the caller.  (This limitation is described more
    precisely in the following section.)
 2. The behavior of recursive cross-contract calls is undefined.
-3. Contracts to be called by other contracts must not define witness
+3. Circuits called from other contracts must not rely on witness
```
dybvig, line 100: It would be nice to state here for the record that private data can be passed into and returned from a child contract via the parent contract, so that this is more of an inconvenience than a limitation.

coips/coip-0002.md
```diff
@@ -200,47 +201,48 @@ also implements the `Adder` interface.
...
-interface by exporting definitions with matching types and purity
-declarations for all the interface's declared circuits, no matter
+To reiterate what was previously stated, a contract satisfies a
+contract type by exporting definitions with matching types and purity
+declarations for all the type's declared circuits, no matter
 whether a `contract implements` declaration appears in the code.  If
 such a declaration _is_ present, though, then the compiler will verify
 it and reject the program with a compile-time error if the contract
-fails to export the circuits required by the interface.
+fails to export the circuits required by the contract type.
```
dybvig: Maybe say explicitly that the contract can export other circuits not listed in the contract type.

coips/coip-0002.md
```diff
@@ -316,73 +325,87 @@ discovery of the circuit code for other contracts.  To make it
...
-3. A contract that declares witnesses is unable to implement any
-   interface.  This implies that every contract reference is to a
-   contract with no private state, and every cross-contract call
-   produces an empty private-state transcript.
+   a contract type `T`.  If the application uses an implementation of
+   `T` in which the called circuit itself makes another call to a
+   circuit defined in a contract type `S`, the compiler may reject the
+   program, or the call may fail at execution time in the deployed
+   application.
+3. Any circuit that calls witness functions is excluded from
```
dybvig, line 352: Let's remember to implement this restriction with a static check in the compiler.

coips/coip-0002.md
```diff
@@ -316,73 +325,87 @@ discovery of the circuit code for other contracts.  To make it
...
+   consideration for satisfying a contract type.  This implies that
+   every contract reference is to a contract that does not manipulate
+   private state in the circuits that can be called through the
+   reference.  The called contract may define witness functions, but
+   those functions cannot be called directly or indirectly by the
+   exported circuits that satisfy a contract type.
 
 Taken together, these limit Compact's current multi-contract systems
```
dybvig, line 361: Again, we should mention that limitation 3 is more of an inconvenience than a limitation.

dybvig (changes requested):

coips/coip-0002.md
```diff
@@ -316,73 +325,87 @@ discovery of the circuit code for other contracts.  To make it
...
-keywords in Compact.  The grammar production for interfaces replaces
-the existing one that used the keyword `contract`, but the contract
-type declaration form was not previously specified to have any
-meaning.
+This proposal introduces new uses for the keywords `contract` and
+`implements`.  Both of these were already reserved keywords in Compact.
+The production for `contract` was already present in the Compact
+grammar, but it was not previously specified to have any meaning.
 
 Thus, this is not a breaking change.  No working Compact programs
```
dybvig, line 416: At the language level, yes.  But in our implementation, the generated JavaScript code is not backward compatible, so if anyone is using that directly, at least, they will have to adapt.  I don't know whether MN.js still provides the same interface to DApps.

coips/coip-0002.md
jonathan-sobel: Done.

coips/coip-0002.md
jonathan-sobel: Done.

coips/coip-0002.md
jonathan-sobel: Done.

coips/coip-0002.md
jonathan-sobel: Clarified.

coips/coip-0002.md
jonathan-sobel: The code in `analysis-passes.ss` does not appear to implement this check currently, at least when checking the `contract implements` declaration.

dybvig: That's correct; it does not.

dybvig (approved):
LGTM

PR #544, opened 2026-06-29
refactor: secp scalar and base field representation

golddydev:


compiler/analysis-passes.ss
```diff
@@ -2304,14 +2304,14 @@
          (source-errorf src "invalid type ~a for witness ~a return value:\n  witness return values cannot include contract values"
                         (format-type type)
                         (id-sym function-name)))
-       #;(when (and (not (feature-zkir-v3)) (contains-secp256k1? type))
+       (when (and (not (feature-zkir-v3)) (contains-secp256k1? type))
```
kmillikin, line 2254: I wasn't aware that we had commented this out!

compiler/typescript-passes.ss
```diff
@@ -2869,7 +2869,7 @@
                   [("Uint8Array") "new Uint8Array(0)"]
                   [("JubjubPoint") "({x: 0n, y: 1n})"]
                   ;; TODO(kmillikin): use a real default for Secp256k1Point.
-                  [("Secp256k1Point") "({x: 1n, y: 29896722852569046015560700294576055776214335159245303116488692907525646231534n})"]
+                  [("Secp256k1Point") "({x: 1n, y: 29896722852569046015560700294576055776214335159245303116488692907525646231534n, identity: false})"]
```
kmillikin: Should be: {x: 0n, y: 0n, identity: true}.  I will make that change.

flake.nix
```diff
@@ -31,7 +31,7 @@
       # NOTE: if this is an internal release (uses -alpha, -beta, or -rc) do NOT update the package.json in runtime
       # since npm can only access public releases. For the compact-runtime release nix will pull in the correct
       # version from this url.
-      url = "github:midnightntwrk/midnight-ledger/ledger-9.1.0.0-rc.3"; # zkir-v2
+      url = "github:midnightntwrk/midnight-ledger/tkerber/zkir-v3-tests"; # zkir-v2
```
kmillikin: There's no reason for this (or the on-chain runtime change).

Trying to keep all the ledger dependencies exactly in lockstep looks like it has caused problems, we can let them diverge.

golddydev: But `ledger-9.1.0.0-rc.3` doesn't expose zkir for secp, does it?

PR #545, opened 2026-06-29
Fix event schema

pataei:
See changelog.

Schema follows:
- https://github.com/LFDT-Minokawa/compact/pull/442/files
- https://github.com/midnightntwrk/midnight-improvement-proposals/blob/main/mips/mip-0002-public-contract-log-emission.md


dybvig (approved):

cosmir17:
Reviewed from the indexer side, the Unshielded event schema matches what the indexer decodes exactly, good to go for us.

- UnshieldedSpend / UnshieldedReceive (145): sender/recipient Either(65) + domainSep(32) + tokenType(32) + amount(16). The added domainSep lands right after the Either, which is where our decoder reads it (offset 65), and the field order matches.
- UnshieldedBurn (113): Either(65) + tokenType(32) + amount(16), no domainSep, matches.
- UnshieldedMint (80): domainSep(32) + tokenType(32) + amount(16), no sender, matches.
- Either stays 65 bytes (is_left + left + right), the sizing our #1279 fix aligned to.

The renames (domain_sep to domainSep, token_type to tokenType, contract_address to contractAddress) don't affect the indexer, we decode positionally by offset not by name, and our GraphQL field names are our own. exports.md mirrors the .ss, consistent.

Net, once this ships in a compact RC our Spend/Receive decode lights up with no indexer change, they already expect 145. LGTM from the indexer side.

PR #561, opened 2026-06-30
Interpret message hash bytes as a big-endian integer

iquerejeta:
The ECDSA functions internally reverse the msg hash, before converting it to a SECP256k1 scalar.


compiler/zkir-v3-library.compact
```diff
@@ -21,67 +21,53 @@ export struct Secp256k1EcdsaSignature {
...
-// expensive in-circuit square root (liftX). The circuit verifies R.x == r as bytes.
-export struct Secp256k1EcdsaSignatureWithRecovery {
-  r: Secp256k1Scalar,
-  s: Secp256k1Scalar,
-  R: Secp256k1Point,
+// Deserialize a 32-byte hash digest into a secp256k1 scalar.
+//
+// By ECDSA convention (https://www.rfc-editor.org/rfc/rfc6979#section-2.3.2) the 
+// digest is the big-endian encoding of an integer, which is then reduced the 
+// scalar field's order.
```
miguel-ambrona:
```suggestion
// digest is the big-endian encoding of an integer, which is then reduced modulo
// the scalar field's order.
```

compiler/zkir-v3-library.compact
```diff
@@ -21,67 +21,53 @@ export struct Secp256k1EcdsaSignature {
...
+circuit hashToSecp256k1Scalar(digest: Bytes<32>): Secp256k1Scalar {
+  const beReversed: Bytes<32> = Bytes[
+    digest[31], digest[30], digest[29], digest[28], digest[27], digest[26], digest[25], digest[24],
+    digest[23], digest[22], digest[21], digest[20], digest[19], digest[18], digest[17], digest[16],
+    digest[15], digest[14], digest[13], digest[12], digest[11], digest[10], digest[9],  digest[8],
+    digest[7],  digest[6],  digest[5],  digest[4],  digest[3],  digest[2],  digest[1],  digest[0]];
```
miguel-ambrona: Seriously? 🤕

tests-e2e/src/tests/compiler/compiler.ecdsa.e2e.test.ts
```diff
@@ -23,14 +23,14 @@ import { sha256 } from '@noble/hashes/sha2.js';
 // A Secp256k1Point is an affine point; a Secp256k1Scalar is a bare bigint.
 type Point = { x: bigint; y: bigint };
 type Signature = { r: bigint; s: bigint };
-type SignatureWithRecovery = Signature & { R: Point };
+// Ethereum-style recoverable signature: r, s, and the recovery id v.
+type RecoverableSignature = Signature & { recovery: number };
```
miguel-ambrona: Is "recoverable signature" a name you made up or is it standard?

IIUC, what you recover is the pk, not the signature.

A better name would be `PkRecoverableSignature`.

miguel-ambrona (approved):

compiler/zkir-v3-library.compact
kmillikin: Unfortunately, I think so for now.  This compiles into 64 `div_mod_power_of_two` instructions and 30 `reconstitute_field` instructions (?!).

miguel-ambrona: Those are really expensive ZKIR instructions, even after this optimization:
https://github.com/midnightntwrk/midnight-ledger/pull/154

Shall we add a rev instruction on `Bytes32`?

kmillikin: Probably, we'd have to expose a `reverse` circuit in the standard library.

@iquerejeta you will use only half as many `div_mod_power_of_two` instructions if you instead cast to a vector:

```
  const v = digest as Vector<32, Uint<8>>;
  const beReversed = Bytes[
    v[31], v[30], v[29], v[28], v[27], v[26], v[25], v[24],
    v[23], v[22], v[21], v[20], v[19], v[18], v[17], v[16],
    v[15], v[14], v[13], v[12], v[11], v[10], v[9],  v[8],
    v[7],  v[6],  v[5],  v[4],  v[3],  v[2],  v[1],  v[0]
  ];
```

kmillikin (approved):

PR #578, opened 2026-07-02
Add a toBinaryRepr for Compact types that replicates the ledger's

kmillikin:
The ledger's Rust code has a `binary_repr` function which doesn't appear to be exposed in the on-chain runtime.   We need an equivalent, for example, for Keccak-256 hashing in the Compact runtime to match the hashing performed in the ZKIR implementation.


runtime/src/compact-types.ts
```diff
@@ -647,3 +647,50 @@ export const ShieldedCoinRecipientDescriptor = {
...
+  const ocrtValue = rtType.toValue(value);
+  const alignment = rtType.alignment();
+
+  // 1. Accumulate Uint8Array pieces.
+  const arrays = [];
+  let length = 0;
+  for (let i = 0; i < alignment.length; ++i) {
+    let segment = alignment[i];
+    if (segment.tag != 'atom') {
+      // We are decoding our own FAB representation and we don't use 'option' so we shouldn't see it
```
dybvig: don't use `option` or don't use `atom`?

CHANGELOG.md
```diff
@@ -5,6 +5,27 @@ All notable changes to this project will be documented in this file.
...
+
+## Fixed
+
+- Add a `toBinaryRepr` to the Compact runtime that replicates the effect of the
+  on-chain Rust `binary_repr`.  Use it in the runtime for the argument to the
+  Noble hashes `keccak_256` function, to correctly replicate the in-circuit
+  implementation.  The ensures that trailing zero bytes from byte vectors are
+  preserved and hashed in JS as well as in circuit.
+
+- Change casting of byte vectors to foreign fields so that the perform modular
```
dybvig: "the" => "they"

dybvig (approved):
LGTM.  Just one change-log typo and question about a comment.

runtime/src/compact-types.ts
kmillikin: I guess I should say "we only use 'atom'".  (There are only two possibilities, 'option' and 'atom'.)

PR #589, opened 2026-07-06
fix: give enums a one-byte minimum width

### LINKED ISSUE #588: persistentHash of a struct with a 0-bit field (single-variant enum / Uint{maxval:0}) emits inconsistent zkir
### Context & versions:

Reproduced with compactc 0.31.104, runtime-version 0.16.101.

A struct field whose FAB encoding is 0 bits wide makes `persistentHash<Struct>(...)` emit a self-inconsistent `persistent_hash` instruction in the zkir: the instruction still receives an input element for the field, but the declared alignment records the field as `Bytes<0>`, which contributes zero elements. The inputs therefore carry one more element than the alignment declares, and every conformant prover rejects the circuit, so any circuit that reaches the hash is unprovable.

The natural way to hit this is a single-variant enum, which lowers to a 0-bit unsigned integer (maxval 0). Any other 0-bit field triggers it the same way.

### Failing example

```compact

enum Kind { mint }   // single variant, lowers to a 0-bit Uint

struct Attestation {
  kind: Kind,
  amount: Uint<64>,
}

export ledger digest: Bytes<32>;

export circuit attest(amount: Uint<64>): [] {
  const h = persistentHash<Attestation>(Attestation { kind: Kind.mint, amount: amount });
  digest = disclose(h);
}
```

The inconsistency is visible directly in the emitted `zkir/attest.zkir`, no proving needed. The instruction receives two input elements while the alignment declares one (the `length: 0` atom contributes zero):

```json
{ "op": "persistent_hash",
  "alignment": [
    { "tag": "atom", "value": { "tag": "bytes", "length": 0 } },
    { "tag": "atom", "value": { "tag": "bytes", "length": 8 } }
  ],
  "inputs": [2, 0] }
```

Proving any circuit that reaches this hash fails, on both the local Rust prover (midnight-zkir 2.1.0) and midnightntwrk/proof-server:8.1.0:

```
midnight_zkir::ir_vm: Inputs did not match alignment
```

### Working example

Give the field a non-zero width, for example a second enum variant, and the same circuit is consistent:

```compact
enum Kind { mint, burn }   // 1 byte
```

```json
{ "op": "persistent_hash",
  "alignment": [
    { "tag": "atom", "value": { "tag": "bytes", "length": 1 } },
    { "tag": "atom", "value": { "tag": "bytes", "length": 8 } }
  ],
  "inputs": [2, 0] }
```

### Expected behavior:

The `persistent_hash` inputs and the declared alignment must agree for a 0-bit field: either the alignment carries an atom for the element that is pushed, or no element is pushed at all.

RomarQ:
Closes https://github.com/LFDT-Minokawa/compact/issues/588

A single-variant enum lowered to Uint<0..0>, a zero-bit type encoded as Bytes<0>. That had two consequences:

- Hashing a struct containing one emitted a self-inconsistent persistent_hash gate: the circuit flattening still allocates one element for the field, so the gate received one more input than its declared alignment accounts for and every conformant prover rejected it (midnight_zkir: "Inputs did not match alignment"). transient_hash and hash_to_curve carry no alignment in the gate, so they silently computed a different hash than the runtime.

- The layout was unstable on the natural upgrade path: enums are extension points, and the width jumped from 0 to 1 bytes the moment a second variant was added, changing hashes, state layouts, and derived addresses. The width rule is otherwise stable across 2 to 256 variants; single-variant was the lone anomaly, and its discriminant was cryptographically silent in every hash.

Lower single-variant enums to the same one-byte unsigned type as two-variant enums (an inclusive upper bound of 1 instead of 0) in replace-enums and give the TS descriptor the same one-byte minimum (CompactTypeEnum keeps the true variant bound for input validation). These are the only two places the compiler computes enum width, so zkir alignments, contract state encodings, and runtime hashing all follow. Regression goldens in both zkir backends pin the struct-field case and the bare-enum case.

## Before

| variants | maxval | width |
| --- | --- | --- |
| 1 | 0 | 0 bytes |
| 2 to 256 | 1 to 255 | 1 byte |

## After

| variants | maxval | width |
| --- | --- | --- |
| 1 to 256 | 1 to 255 | 1 byte |


RomarQ:
Closing in favor of https://github.com/LFDT-Minokawa/compact/pull/622

PR #592, opened 2026-07-06
fix(events): serialize ShieldedReceive fields in the CoIP-442/MIP-0002 order

### LINKED ISSUE #590: midnight-events.ss: ShieldedReceive field order does not match CoIP-442 / MIP-0002
## Summary

`compiler/midnight-events.ss` declares `ShieldedReceive` as `(commitment, contractAddress, ciphertext)`, but the agreed order is `(commitment, ciphertext, contractAddress)`: that is what MIP-0002 Appendix A (midnightntwrk/midnight-improvement-proposals, main) specifies, and this repo's own CoIP-442 was reordered to match in e537fc9 ("feat(coip,events): Reorder ShieldedReceive fields"). The implementation was never updated, so the compiler contradicts its own COIP. Present on the `v0.33.0-rc.0` tag and on `main`.

## Impact

Consumers that decode the agreed order (the midnight-indexer does) mis-decode `ciphertext` and `contractAddress` for every ShieldedReceive emitted by 0.33.0-rc.0 — silently, with plausible-looking values (the two `Maybe` fields are read at each other's offsets). Byte-exact analysis and reproduction: midnightntwrk/midnight-indexer#1311 (repro harness: midnightntwrk/compact-end-2-end#46).

## Fix

Swap the two fields in the `ShieldedReceive` `declare-event-type` (sizes unchanged, total stays 578) and align `doc/api/CompactStandardLibrary/exports.md`. Same catch-up shape as the domainSep alignment in #545.

cosmir17:
## Summary

`compiler/midnight-events.ss` declares `ShieldedReceive` as `(commitment, contractAddress, ciphertext)`, while CoIP-442 (since e537fc9, "Reorder ShieldedReceive fields") and MIP-0002 Appendix A specify `(commitment, ciphertext, contractAddress)`. Emitted events therefore don't match consumers that decode the specified order — the midnight-indexer does, which is how this surfaced (midnightntwrk/midnight-indexer#1311: `ciphertext` and `receivingContractAddress` mis-decode, byte-exact repro). The field order lives in exactly two places across the stack, this declaration and the indexer's decoder; compact-runtime, compact-js, the node toolkit and midnight.js carry no field-order knowledge, so no other component needs a code change.

## Changes

- `compiler/midnight-events.ss`: swap the two `Maybe` fields in the `ShieldedReceive` declaration (serialized size unchanged, 578).
- `doc/api/CompactStandardLibrary/exports.md`: align the `ShieldedReceive` struct doc.
- `compiler/test.ss`: update the six `tstruct ShieldedReceive` expectation fixtures to the new field order, and reorder the fields in the four `emit(ShieldedReceive { … })` test-source literals to match the declaration.
- `CHANGELOG.md`: entry under the unreleased toolchain 0.33.100 section.

## Notes for reviewers

- The expectation fixtures were updated structurally (field forms swapped, parenthesis balance verified); I couldn't run the compiler test suite locally, so please let CI validate, and if any expectation is pretty-printer-formatting-sensitive it may need regeneration.
- The JS-side test objects (`{ commitment: …, contractAddress: …, ciphertext: … }` in staged JS) are field-name keyed and were left as is.

Fixes #590.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Toolchain 0.33.100, language 0.25.100, runtime 0.18.0]
 
+### Fixed
+
+- The `ShieldedReceive` standard event now serializes its fields in the order
+  specified by CoIP-442 and MIP-0002: `commitment`, `ciphertext`,
+  `contractAddress` (previously `contractAddress` preceded `ciphertext`).
+  Serialized size is unchanged (578). Fixes #590.
+
 ### Changed
 
 - The standard library ECDSA circuits `secp256k1EcdsaVerify` and
--- a/compiler/midnight-events.ss
+++ b/compiler/midnight-events.ss
@@ -20,8 +20,8 @@
 (declare-event-type ShieldedReceive 1 578
   "A contract accepts an incoming shielded coin.\n`contractAddress` set when received by a contract, absent for user recipients."
   ([commitment (Bytes 32) (hint "indexed")]
-   [contractAddress (TypeRef Maybe (Bytes 32))]
-   [ciphertext (TypeRef Maybe (Bytes 512))]))
+   [ciphertext (TypeRef Maybe (Bytes 512))]
+   [contractAddress (TypeRef Maybe (Bytes 32))]))
 
 (declare-event-type ShieldedMint 2 81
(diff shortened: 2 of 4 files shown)


JosephDenman (approved):

kmillikin:
It's relatively minor, but we should have bumped the version numbers correctly instead of piggy-backing off an existing one.

PR #598, opened 2026-07-07
Bump versions

JosephDenman:


flake.nix
```diff
@@ -31,21 +31,21 @@
...
       # version from this url.
-      url = "github:midnightntwrk/midnight-ledger/ledger-9.1.0.0-rc.2"; # zkir-v2
+      url = "github:midnightntwrk/midnight-ledger/ledger-9.1.0.0-rc.3"; # zkir-v2
       inputs.zkir.follows = "zkir";
     };
     onchain-runtime-v4 = {
       # dependency for compact-runtime release
       # all notes for the zkir input applies to onchain-runtime input too.
-      # NOTE: ledger-9.1.0.0-rc.2 is the first tag packaging the wasm under the
+      # NOTE: ledger-9.1.0.0-rc.3 is the first tag packaging the wasm under the
```
kmillikin: I think this note is probably not correct.  I think -rc.1 is the first such tag judging by https://www.npmjs.com/package/@midnightntwrk/onchain-runtime-v4?activeTab=versions

But I wouldn't change the comment here anyway.  We're going to merge back to `main` at the end of the release cycle and we want to minimize merge conflicts.

(The comment seems of dubious value, not really the right place to document the past history of this file.)

flake.nix
```diff
@@ -31,21 +31,21 @@
...
+      url = "github:midnightntwrk/midnight-ledger/ledger-9.1.0.0-rc.3";
       inputs.zkir.follows = "zkir";
     };
     zkir-wasm = {
       # dependency for test-center
-      url = "github:midnightntwrk/midnight-ledger/ledger-9.1.0.0-rc.2";
+      url = "github:midnightntwrk/midnight-ledger/ledger-9.1.0.0-rc.3";
       inputs.zkir.follows = "zkir";
     };
     zkir-v3 = {
```
kmillikin, line 51: We can also update this to the hash b17df9d100812bfb2621a3e342108158e3c6b412 (https://github.com/midnightntwrk/midnight-ledger/commit/b17df9d100812bfb2621a3e342108158e3c6b412) and the same for `zkir-v3-wasm`.

That matches the what was built for the proof server -rc.5 release that we are telling people to use with ZKIR v3.

JosephDenman: Fixed.

flake.nix
JosephDenman: Fixed.

kmillikin (approved):

PR #600, opened 2026-07-07
Compact reference updates to cover CoIP 2

jonathan-sobel:
Additions and changes to the Compact reference to describe contract types and cross-contract calls.


compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+```
+
+(See later sections of this Compact reference for more information
+about ledger-state types and constructors.)
+
+When some subset of a contract's circuits matches the set of those
+declared in a contract type, the contract is said to **satisfy** (or
+**have**) that type.
+
+More precisely, define the **signature** of a circuit to be the types
```
dybvig: I couldn't parse this at first.  Perhaps rewrite this as `the **signature** of a circuit consists of type types ...`

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+of its parameters, plus its return type, as well as whether it is
+declared to be `pure`.  Changing the parameter names or using
+destructuring patterns in a circuit definition does not change its
+signature, because the signature is defined only by the parameter and
```
dybvig: `defined` => `determined` is better with the change suggested above.

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+return *types*.
+
+The circuit declarations in a contract type declaration also have
+signatures defined by their types and whether they are declared to be
```
dybvig: `defined` => `determined`

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+pure.  For example, the signature of `addTo` in the `Adder` interface
+above is
+
+> `Uint<64>` &rarr; `Uint<64>`
+
+A contract **satisfies** an contract type when, for every circuit
```
dybvig: `an` => `a`

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+circuits with signatures identical to those in the `Adder` declaration
+above satisfies the `Adder` contract type.  A contract whose
+definition of `addTo` is declared to be `pure` would also satisfy the
+`Adder` type.
+
+This is true regardless of whether the contract declares any intention
+to satisfy the type.  In other words, contract typing is structural,
+not nominal.
+
+On the other hand, when a contract type used to describe the required
```
dybvig: `used` => `is used`

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+```
+
+The contract type named in the `contract implements` form must be in
+scope, using Compact's existing rules for the scope of program-defined
+types.  The `contract implements` declaration applies to the contract
+being defined at the point where it appears.  A contract may be
+declared to satisfy several different contract types by including
+multiple `contract implements` declarations.
+
+A contract type represents the set of deployed contracts that satisfy
```
dybvig: `set of` => `set of all possible`

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+the corresponding contract type definition.  A contract value (i.e., a
+value with a contract type) might be represented by the address of the
```
dybvig: `the deployed contract` => `a deployed contract`

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+deployed contract.
+
+No mechanism is provided within the Compact language to *create*
+values with contract types.  Instead, the surrounding context (that
+is, the application code) introduces a contract value into the Compact
+program by calling a circuit or constructor with the address of a
```
dybvig: Above you said it _might_ be represented as an address; here you imply it _is_.

compiler/compact-reference-proto.mdx
```diff
@@ -617,10 +617,145 @@ names (in the same order) and distinct otherwise.
...
+deployed contract or by returning the address of a deployed contract
+from a witness function.
+
+To reiterate what was previously stated, a contract satisfies a
+contract type by exporting definitions with matching types and purity
+declarations for all the type's declared circuits, no matter whether a
+`contract implements` declaration appears in the code.  If such a
+declaration _is_ present, though, then the compiler will verify it and
```
dybvig: eliminate `then`.

compiler/compact-reference-proto.mdx
```diff
@@ -2586,6 +2741,11 @@ named circuit.
 When *id* refers to a generic witness or circuit, it must be fully specialized via
 generic parameters.
 
+To make a cross-contract call to a circuit in another contract, *fun*
```
dybvig: In the current grammar, a `Function` cannot take the form *e*` . `*id*.  Instead, cross-contract calls are recognized by the third production of `expr8`, which is labeled "element call" in `parser.ss`.  This is the same production that recognizes calls to ledger operators, e.g., `mylist.insert(x)`.  We should probably experiment with allowing `Function` to take the form *e*` . `*id* and eliminating the third `expr8` production.  If that works, we can leave this here and give a forward reference to the section on ledger-state operations.

Otherwise, we should probably address cross-contract calls in a separate section, as we do for ledger-state operations.

compiler/compact-reference-proto.mdx
```diff
@@ -3306,6 +3490,38 @@ from looping indefinitely on programs that operate on excessively large vector
...
+determining whether a contract satisfies a contract type, the current
+compiler eliminates from consideration any circuits that call
+witnesses.  A called contract may define witness functions, but
+those functions cannot be called directly or indirectly by the
+exported circuits that satisfy a contract type.  This implies that
+every cross-contract call is to a circuit whose only access to private
+state is through its arguments, and its only means of affecting
+private state is through the values it returns.
+
+Furthermore, In order to satisfy a contract type `T`, a contract must
```
dybvig: I don't believe the first part of this is true, i.e., that the contract need be defined by `T.compact`.  I believe the only requirement is that the JavaScript code that implements the contract be present in the file `../T/contract/index.js`.

dybvig (changes requested):

compiler/compact-reference-proto.mdx
dybvig: Actually, changing the grammar is probably not going to work for a few reasons:
1. We probably don't want to allow e.id to appear in the function position of a `map` or `fold`.
2. In the most natural expression, were `e` is an `expr`, we'd end up parsing a ? b : c.id(...) as  `(a ? b : c).id(...)` rather than as `a ? b : (c.id(...))`.   We'd have to use `expr8` in an attempt to get the same parse as before, which is a little ugly.
3. I don't think even `expr8` would work here because Calls is at `expr8` (binds more tightly).

jonathan-sobel: *Logically*, the thing in the operator position in a circuit call can be `e.id`, even if the parsing is handled by a different part of the grammar than the other function-call forms. I think it makes sense to document them together. All the type-checking rules for calls are already in the section on circuit and witness calls.

Let me try to work on it some more, and if I can't make it work, we'll create a new section.

Oh, I do have a question, though: does the compiler handle overloading in contract types and cross-contract calls? That is, can I define a contract type containing two different circuits named `c`, as long as they have different numbers/types of parameters? And does the right thing happen at the call site?

compiler/compact-reference-proto.mdx
jonathan-sobel: Done.

compiler/compact-reference-proto.mdx
jonathan-sobel: Done.

compiler/compact-reference-proto.mdx
jonathan-sobel: Done.

compiler/compact-reference-proto.mdx
jonathan-sobel: Done.

compiler/compact-reference-proto.mdx
jonathan-sobel: Done.

compiler/compact-reference-proto.mdx
jonathan-sobel: Done.

compiler/compact-reference-proto.mdx
jonathan-sobel: Good catch. Here's the new version: 
> ... introduces a contract value into the Compact program by calling a circuit or constructor with a value representing a deployed contract or by returning a contract value from a witness function.

compiler/compact-reference-proto.mdx
jonathan-sobel: Done.

compiler/compact-reference-proto.mdx
jonathan-sobel: You're right. Since this is the specific documentation for what's implemented, and not just the proposal, we can be very concrete about what's required. Rewritten. I think it's much better now.

compiler/compact-reference-proto.mdx
jonathan-sobel: Reworked the parts about cross-contract calls in the "Circuit and witness calls" section. I think it address the concerns, while keeping the information about calls together.

Also, I understand now that, because contract types can refer only to circuits exported from the top level, and it's not possible to export two circuits with the same name from the top level, there is no overloading in contract types and cross-contract calls.

compiler/compact-reference-proto.mdx
dybvig: I still think any type represents the set of all things that satisfy that type.  We wouldn't say that `Uint<8>` represents only the values actually used in a program; rather, we'd say it represents all possible values in the range 0..255 inclusive.

compiler/compact-reference-proto.mdx
```diff
@@ -2759,6 +2754,17 @@ Finally, *fun* can be a parenthesized function `(`*fun*`)`.
 Because circuits and witnesses are not first class, *fun* cannot be a variable
 name or arbitrary expression.
 
+**Cross-contract calls:** Circuits named in contract types are called
```
dybvig: To clarify, my original problem leaving this here is that someone looking at what a *fun* can be above might be confused to find that *fun* can also be *e.id*.  Reiterating the relevant *expr8* production here helps a little, but not enough IMO.  As I read this through again, I'm also thinking that cross-contract calls are inherently special, like ledger-state operations, and deserve their own section that puts all of the information about them together in one place, including current limitations.

compiler/compact-reference-proto.mdx
```diff
@@ -3494,22 +3501,27 @@ a `for` loop.
...
+may be called by other circuits, but they cannot be called directly or
+indirectly by the exported circuits that satisfy a contract type.
+This implies that every cross-contract call is to a circuit whose only
+access to private state is through its arguments, and its only means
+of affecting private state is through the values it returns.
+
+Furthermore, the current implementation of cross-contract calls
+provides no means of discovering any code for called circuits, other
+than the code provided by the calling application.  What this
+generally means is that an application satisfies a contract type `T`
```
dybvig: "generally" => "typically"

"compiling a contract" => "compiling a specific contract"

The presentation of this restriction seems backward.  I would move the sentence "The code that is generated ... expects to find" before "What this generally [typically] means ...".

dybvig (changes requested):

compiler/compact-reference-proto.mdx
```diff
@@ -3306,6 +3509,43 @@ from looping indefinitely on programs that operate on excessively large vector
...
+**Limitations on contract types and cross-contract calls:**
+When computing the set of exported circuits for the purpose of
+determining whether a contract satisfies a contract type, the current
+compiler eliminates from consideration any circuits that call
+witnesses.  A called contract may define witness functions, and they
+may be called by other circuits, but they cannot be called directly or
+indirectly by the exported circuits that satisfy a contract type.
+This implies that every cross-contract call is to a circuit whose only
+access to private state is through its arguments, and its only means
+of affecting private state is through the values it returns.
```
JosephDenman: I would say, "This implies that every cross-contract call is to a circuit whose only access to private data is through its arguments." Or better yet remove the "This implies..." sentence entirely. _We_ know there's a transform from programs with witnesses to programs without them by encoding private state in arguments and return values. We cannot assume that will be obvious to our users, and I think it's not necessary to describe the feature as is.

compiler/compact-reference-proto.mdx
```diff
@@ -3306,6 +3509,43 @@ from looping indefinitely on programs that operate on excessively large vector
...
+
+Furthermore, the current implementation of cross-contract calls
+provides no means of discovering any code for called circuits, other
+than the code provided by the calling application.  What this
```
JosephDenman:
```suggestion
than the code provided by the application.  What this
```

compiler/compact-reference-proto.mdx
```diff
@@ -3306,6 +3509,43 @@ from looping indefinitely on programs that operate on excessively large vector
...
+generally means is that an application satisfies a contract type `T`
```
JosephDenman:
```suggestion
generally means is that an application provides a contract type `T`
```

I think "satisfies" makes it sound like the application itself has some special role in compilation. There's also a `satisfies` keyword in TypeScript that could plant some spurious associations in the readers head.

compiler/compact-reference-proto.mdx
```diff
@@ -3306,6 +3509,43 @@ from looping indefinitely on programs that operate on excessively large vector
...
+
+The real dependency, however, is not on the source file, but instead
+on the generated JavaScript file.  The code that is generated at the
+site of a call to one of `T`'s circuits expects to find the
+implementation of those circuits in the JavaScript file
+`../T/contract/index.js`.  This effectively limits any DApp to a
+single implementation of each contract type.
+
+If any cycles are present in the call graph of cross-contract calls,
+the behavior is undefined.  This implies that contracts cannot
```
JosephDenman: It's more than undefined. It's erroneous. The runtime will throw an error that a re-entrancy has occurred.

compiler/compact-reference-proto.mdx
jonathan-sobel: Agree. Fixing.

JosephDenman (changes requested):

compiler/compact-reference-proto.mdx
jonathan-sobel: I'm following the recommendation to remove the bit about the return value, but I think it's OK to leave in the statement about access to private state, without implying that it has anything to do with a program transformation.

JosephDenman: It's specifically the "state" phrasing that bothers me because "witness state" and "ledger state" are already terms we use everywhere.

compiler/compact-reference-proto.mdx
jonathan-sobel: I think it's important to leave "calling" in, because an application associated with the called contract (e.g., imagine a DApp for a token) cannot provide the code, only the application making the call.

compiler/compact-reference-proto.mdx
jonathan-sobel: I meant "satisfy" in the technical sense that's defined in the section on contract types, but I see how it's unclear here. It's better if it's a little more verbose: "... is that an application provides circuits that satisfy a contract type `T` by ..."

compiler/compact-reference-proto.mdx
jonathan-sobel: For one thing, I wasn't sure we could guarantee that in all cases. But more importantly, it's standard practice to label something like this as "undefined behavior" in a language specification, because otherwise, the error is part of the defined behavior of programs, and it's a non-backwards-compatible change to eliminate the restriction. In other words, we don't want to say that a program containing cross-contract recursion is well-defined to throw an error; we want to say that the beahvior is unspecified, so that we can specify it later.

I'm willing to be overruled here, if other maintainers think differently.

compiler/compact-reference-proto.mdx
jonathan-sobel: Further improved: "... every cross-contract call is to a circuit whose only access to private data is through the values given as its arguments."

JosephDenman (approved):
LGTM.

dybvig (approved):

PR #604, opened 2026-07-10
Reject secp256k1 identity points in Ethereum address computation

iquerejeta:
This PR rejects identity points from the ethereum address computation function.


compiler/test.ss
```diff
@@ -50170,110 +50170,110 @@ groups than for single tests.
...
         (public-ledger-declaration
-          (%x0.13 (__compact_Cell (tfield (field-native))))
-          (%x1.14 (Set (tfield (field-native))))
-          (%x2.15 (Counter))
-          (%x3.16 (List (tfield (field-native))))
-          (%x4.17 (Map (tfield (field-native)) (tboolean)))
-          (%x5.18 (MerkleTree 32 (tfield (field-native))))
-          (%x6.19 (HistoricMerkleTree 10 (tfield (field-native))))
-          (%x7.20
+          (%x0.28 (__compact_Cell (tfield (field-native))))
```
kmillikin, line 50173: I don't know why these aren't stable.  I just regenerated the expectations.

tests-e2e/src/tests/compiler/compiler.ecdsa.e2e.test.ts
```diff
@@ -105,31 +89,38 @@ describe('[ECDSA] examples/ecdsa/example_one.compact', () => {
         expect(recovered.y).toBe(pk.y);
     });
 
-    test('ethereumAddress hashes a recovered public key to 20 bytes [WIP gates]', () => {
+    test('secp256k1EthereumAddress hashes a recovered public key to 20 bytes [WIP gates]', () => {
```
JosephDenman, line 92: Is there a reason why we aren't asserting that a well known Ethereum address can be recovered here? I don't see any actual tests that an Ethereum address can actually be recovered.

JosephDenman (commented):
Just one main question really.

tests-e2e/src/tests/compiler/compiler.ecdsa.e2e.test.ts
kmillikin: I think we have tests in the end-to-end repository like that (at least @golddydev has reported issues).  But the basic correctness is not really an end-to-end matter.

I've added a test and fixed a pair of issues that were revealed.  I used https://www.rfctools.com/ethereum-address-test-tool/ to generate the expectation from a random private key (secp256k1 scalar) gotten by pressing the button on that page.

compiler/test.ss
dybvig: The identifiers are assigned names the first time they are printed so they will be consistent across all of the output forms.  Probably the `pass-returns` form for `infer-types` was added after the `pass-returns` form for `reject-recursive-circuits` and caused the latter's names to be out-of-sequence.  Since the checker uses extensional equivalence wrt names, the issue would not arise until there was some other difference in the generated output, or one that wasn't corrected proactively by hand, as I often do.

dybvig (approved):
LGTM.  Just a response to your question about stability of names in the `pass-returns` output.

JosephDenman (approved):
LGTM

PR #606, opened 2026-07-10
Implement proper Secp256k1Point equality

kmillikin:
The code for Secp256k1Point equality is:

```javascript
_equal_0(x0, y0) {
  if (x0.identity) { return y0.identity; }
  if (y0.identity || x0.x != y0.x || x0.y != y0.y) {
    return false;
  }
  return true;
}
```

Also change JubjubPoint equality to the more simple:

```javascript
_equal_0(x0, y0) {
  if (x0.x != y0.x || x0.y != y0.y) {
    return false;
  }
  return true;
}
```

And add `Uint` types as ones that have simple JS `===` equality.


compiler/language-version.ss
```diff
@@ -15,11 +13,13 @@
...
 ;;; limitations under the License.
 
+#!chezscheme
+
 (library (language-version)
   (export language-version-string check-language-version)
   (import (chezscheme) (version))
 
-  (define language-version (make-version 'language 0 25 101))
+  (define language-version (make-version 'language 0 25 102))
```
dybvig, line 22: I don't think this is a language change, but we should decide what is.  For now, let's go with what you have.

compiler/typescript-passes.ss
```diff
@@ -3066,7 +3071,7 @@
                0 (Expr expr2 (precedence add1 ==) outer-pure?)))
            (parenthesize level (precedence call)
              (let ([equal-name (unique-global-name "equal")])
-               (build-equal-helper! type equal-name)
+               (build-equal-helper! (de-alias type) equal-name)
```
dybvig: The de-alias of type is already performed within `build-equal-body`.

compiler/typescript-passes.ss
```diff
@@ -3089,7 +3097,7 @@
                0 (Expr expr2 (precedence add1 ==) outer-pure?)))
            (parenthesize level (precedence not)
              (let ([equal-name (unique-global-name "equal")])
-               (build-equal-helper! type equal-name)
+               (build-equal-helper! (de-alias type) equal-name)
```
dybvig: Ditto.  The de-alias of type is already performed within `build-equal-body`.

I'm going to go ahead and push a fix for this and the earlier one, plus a test to confirm.

dybvig (changes requested):

dybvig (approved):

PR #612, opened 2026-07-13
Handle alignment for curve points in ZKIR v3

kmillikin:
Curve points (`JubjubPoint` and `Secp256k1Point`) are given a "pseudo-alignment" atom in ZKIR v3.  This is an alignment atom that the ledger doesn't know about and that the ledger is not intended to see.  They were handled specially in the Impact instructions `push` and `popeq`.

However, some ZKIR operations themselves need a proper ledger alignment (`persistentHash`, `persistentCommit`, and `keccak256`) which wasn't handled.


dybvig (approved):
LGTM

PR #613, opened 2026-07-13
Fix a pattern match failure in the circuit optimizer

kmillikin:
A field type comparison predicate did not include the secp256k1 base and scalar fields.


dybvig (approved):
LGTM

PR #622, opened 2026-07-15
Fix issue #588

kmillikin:
The underlying issue is that we have given the type `Uint<0..1>` an alignment of `(abytes 0)` in `flatten-datatypes`.  The ledger expects **no** values for the hashing functions that take an alignment (`persistentHash`, `persistentCommit`, and `keccak256`) and when those values occur in Impact code.  We are providing **one** value (always zero).

When enumeration types are translated away, an `enum` with one variant is represented by a value of type `Uint<0..1>` so it has the same problem.

The fix adopted is to represent these values identically with the Compact type `Bytes<0>`, as no values after `flatten-datatypes`.  This requires us to materialize a literal 0 in contexts where we do need a value, such as arithmetic and comparison operations.


kmillikin:
As discussed, the alternative is to give `Uint<0..1>` an alignment of `(abytes 1)`.  The ledger will obey us if we tell it that.  It will take up space in the ledger (and contribute a 0 value when hashing).

compiler/circuit-passes.ss
```diff
@@ -2599,17 +2606,24 @@
...
-           (with-output-language (Lflattened Statement)
-             (if (null? triv1*)
-                 (begin
-                   (hashtable-set! var-ht var-name (Wump-single triv-accum))
-                   (list `(= ,test ,var-name ,triv-accum)))
-                 (let ([t1 (make-new-id var-name)] [t2 (make-new-id var-name)])
-                   (cons* `(= ,test ,t1 (== ,(car triv1*) ,(car triv2*)))
-                          `(= ,test ,t2 (select ,triv-accum ,t1 0))
-                          (f (cdr triv1*) (cdr triv2*) t2)))))))]
+         (let-values ([(triv1* triv2*) (if (null? triv1*)
```
dybvig, line 2609: Did you actually see the problem this addresses in practice?  I don't see any new tests involving `==`, but perhaps an existing test broke without this change.

On first glance, I figured the problem shouldn't arise because the operands of `==` have equivalent types and therefore equivalent representations, but they aren't necessarily equivalent types, but merely related types.  (Between `infer-types` and when we drop `safe-cast` they are equivalent, but by now they are back to just being related.)

That is, I think this case arises (only?) when one operand of `==` has a type that contains somewhere within it the type `Uint<0..1>` and the other has a supertype in which  of `Uint<0..1>` is replaced with `Uint<0..k>`, k > 1.  By representing values of type `Uint<0..1>` with zero fields instead of one, we are effectively violating the property that a value of type A that is a subtype of type B has the same representation, hence the need for some adjustment.

This suggests that it would be better to represent values of type `Uint<0..1>` with one field rather than zero.

If we keep the current approach, I'm concerned that we'll have other cases this implementation of `==` doesn't anticipate,  For example, what happens if we compare a value of type `Vector<10, Uint<0..1>>` with a value of type `Vector<10, Uint<0..2>>`?  Or worse, in a way, a value of type `[Uint<0..2>, Uint<0..1>, Uint<0..2>]` with a value of type `[Uint<0..2>, Uint<0..2>, Uint<0..2>]`?

I say the last example is worse because we can see that `triv1* and `triv2* have different lengths, but we don't know where to insert the `0` value.  Maybe we could fix this by including the types in the `==` form and traversing it to find cases where an adjustment is needed, but it doesn't seem worth the complexity.

I'm also not confident that `==` is the end of the problem.

dybvig (changes requested):

compiler/circuit-passes.ss
kmillikin: The specific case where this fails in the existing tests is comparison with a literal.  `circuit test(x: Uint<0..1>): Boolean { return x == 0; }`.

`x` is no values, and `0` is one value so this code materializes a triv 0 to make the lengths equal.  Looking at the implementation of this pass, I see that there's one other case that I missed, not covered by tests, which is `b ? x : n` where `x: Uint<0..1>` and `n` has some `Uint` type.

So on reflection you are right that we should represent all `Uint<0..n>` uniformly with a single native field element in circuit (for now, until we add biguints).  I think then we will have to encode `Uint<0..1>` in the ledger as `(abytes 1)`.

I'll abandon this change and make that one instead.

kmillikin:
I'm abandoning this change, and instead making `Uint<0..1>` represented by one native field value in circuit, using an alignment of `(abytes 1)` in the ledger.

PR #626, opened 2026-07-17
Fix issue #588

kmillikin:
The underlying issue is that we have given the type `Uint<0..1>` an alignment of `(abytes 0)` in `flatten-datatypes` and in the descriptors in the Compact runtime.  The ledger expects **no** values in the transcript for such an alignment, and we are providing **one** native-field value (always zero).  This will result in a proof failure due to the malformed transcript.

When enumeration types are eliminated, `enum`s with one variant are represented by the type `Uint<0..1>` so they share the same problem.

The fix adopted is to give these values an alignment of `(abytes 1)` so that the ledger will expect the value that is provided.


compiler/test.ss
```diff
@@ -89935,11 +90183,167 @@ groups than for single tests.
...
+        "  expect((await contract.circuits.test(context, embedded)).result).toEqual(embedded);"
+        "});"
+        ))
+    )
   )
 
 (run-javascript)
 )
 
+#!eof
```
kmillikin: Oops :)

dybvig (approved):
LGTM

PR #628, opened 2026-07-17
CoIP: Dynamic Selection of Implementation for Cross-Contract Calls

jonathan-sobel:
Remove the limitation defined in CoIP 2, of having a single implementation per contract type. Instead find, load, and run the code associated with each specific contract value.


jonathan-sobel:
@dybvig @kmillikin @JosephDenman I believe the CoIP proposing dynamic dispatch in cross-contract calls is now ready to be reviewed and assigned a number.

kmillikin (approved):

PR #630, opened 2026-07-20
feat: add ecdsa recover builtin

golddydev:
## Summary

This PR is to expose ecdsa recover function which can then be used off-circuit to recovery public key
and feed that to verify in-circuit as either witness or argument.

## Changes

- Add tests to exercise ecdsa recover builtin function
- Add tests under `test.ss` to exercise ecdsa `recover` off-circuit and `verify` in-circuit together

## Notes

- ecdsa `recovery` function doesn't enforce low-S and this is illustrated under `test.ss` as twin signature is passing successfully.


compiler/zkir-v3-library.compact
```diff
@@ -77,6 +77,12 @@ circuit hashToSecp256k1Scalar(digest: Bytes<32>): Secp256k1Scalar {
 //   P  = u1*G + u2*pk
 //   valid iff P.x == r as 32-byte big-endian integers
 //
+// `s` is accepted anywhere in [1, n), both low-s and high-s, as in textbook ECDSA.
```
iquerejeta: cite the digital signature standard instead https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf

runtime/src/built-ins.ts
```diff
@@ -437,6 +433,57 @@ export function secp256k1MulGenerator(b: bigint): Secp256k1Point {
...
+/**
+ * Recover the secp256k1 public key from an ECDSA signature and a message
+ * hash, following Ethereum `ecrecover` semantics.
+ *
+ * The intended pattern is to recover the public key here,
+ * pass it to a circuit as an ordinary (private) argument or
+ * witness, and constrain it in-circuit with `secp256k1EcdsaVerify(msgHash, sig, pk)`.
+ * The recovery itself is untrusted hint computation: only a public key that
+ * actually verifies against (msgHash, sig) will satisfy the in-circuit check.
+ *
```
iquerejeta: This should not be documented here. Let's try to keep the docs as simple as possible. However, could you mention this function in the ecdsa_verify function from compact? Say that if one wants to recover, this function should be called.

runtime/src/built-ins.ts
```diff
@@ -437,6 +433,57 @@ export function secp256k1MulGenerator(b: bigint): Secp256k1Point {
...
+ * the signing key.  (Flipping only one of the two — `s` without the id, or the
+ * id without `s` — recovers a different key, which still verifies.)
+ *
+ * This does not weaken authorisation: the twin recovers the same key, so a
+ * check binding the recovered key to an address the contract already trusts
+ * still holds.  It does matter when a signature is treated as a *unique
+ * identifier*.  Such callers must reject or normalise high-s themselves
+ * (require `s <= n / 2`), or the identifier can be mutated into a second value
+ * that passes every check.
+ */
```
iquerejeta: No need to add this here. We already make a comment in the ECDSA verify function in compact.

compiler/test.ss
```diff
@@ -91029,6 +91030,102 @@ groups than for single tests.
         "});"
         ))
     )
+
+  (test
+    '(
+      "import CompactStandardLibrary;"
+      "export ledger owner: Bytes<20>;"
+      "witness recoveredKey(): Secp256k1Point;"
```
kmillikin: I think we want a witness that actually recovers the key, something with a signature like:

```
witness recoverKey(messageHash: Bytes<32>,
                   signature: Secp256k1EcdsaSignature,
                   signingNonce: Secp256k1Point): Secp256k1Point;
```

I think people both want to know how to assert a recovered key is valid in circuit, and how to recover it in JS code in the first place.

compiler/test.ss
```diff
@@ -91029,6 +91030,102 @@ groups than for single tests.
...
+      "export circuit setOwner(addr: Bytes<20>): [] {"
+      "  owner = disclose(addr);"
+      "}"
+      "// The Ethereum pattern: `require(ecrecover(hash, v, r, s) == owner)`."
+      "// Verification alone does not say *who* signed -- every recoverable id"
+      "// yields a key that verifies -- so the recovered key is additionally bound"
+      "// to an address the contract already trusts."
+      "export circuit verifyOwnerSignature(msg: Bytes<32>,"
+      "                                    sig: Secp256k1EcdsaSignature): [] {"
+      "  const pk = recoveredKey();"
```
kmillikin: And then we might need here:

```
assert(secp256k1PointX(signingNonce) as Bytes<32> as Secp256k1Scalar == sig.r, "some meaningful message");
```

which we had in the circuit that was removed.

compiler/zkir-v3-library.compact
kmillikin: Nobody but us reads comments in the source code.  Text like this belongs in `doc/api/CompactStandardLibrary/exports.md` (this circuit and the argument struct are completely missing from there).  One of us should move this text :)

runtime/src/built-ins.ts
```diff
@@ -437,6 +433,57 @@ export function secp256k1MulGenerator(b: bigint): Secp256k1Point {
   return secp256k1FromProjective(secp256k1.Point.BASE.multiplyUnsafe(b));
 }
 
+/**
+ * Recover the secp256k1 public key from an ECDSA signature and a message
+ * hash, following Ethereum `ecrecover` semantics.
```
iquerejeta: I belive that `ecrecover` returns the Ethereum address. We are just returning the public key. That value then needs to be converted to an ethereum address with the compact function.

iquerejeta (commented):
Looks good - you do mention a few times 'textbook ecdsa'. Can you change that to cite the standard instead?

runtime/src/built-ins.ts
```diff
@@ -437,6 +433,57 @@ export function secp256k1MulGenerator(b: bigint): Secp256k1Point {
...
+ * id without `s` — recovers a different key, which still verifies.)
+ *
+ * This does not weaken authorisation: the twin recovers the same key, so a
+ * check binding the recovered key to an address the contract already trusts
+ * still holds.  It does matter when a signature is treated as a *unique
+ * identifier*.  Such callers must reject or normalise high-s themselves
+ * (require `s <= n / 2`), or the identifier can be mutated into a second value
+ * that passes every check.
+ */
+export function secp256k1EcdsaRecover(msgHash: Uint8Array, sig: Secp256k1EcdsaSignature, recoveryId: number): Secp256k1Point {
```
kmillikin: OK, so this is the witness that we actually want.  A user can just write (and in the test we can use) `const recoverKey = runtime.secp256k1EcdsaRecover`.

However, it does not belong in this file, which contains function definitions that are used in generated Compact code.  It probably goes in `runtime/src/utils.ts`.

runtime/src/compact-types.ts
```diff
@@ -59,6 +59,16 @@ export interface Secp256k1Point {
   readonly identity: boolean;
 }
 
+/**
+ * An ECDSA signature over the foreign secp256k1 elliptic curve. TypeScript
+ * representation of the standard library's `Secp256k1EcdsaSignature` struct.
+ * r and s are secp256k1 scalars in the range [0, SECP256K1_SCALAR_MODULUS).
+ */
+export interface Secp256k1EcdsaSignature {
```
kmillikin: We don't want this.  The compiler will generate its own definition of the struct defined in the standard library, and there is a risk that this one will not match it so it's a hazard to have it here.  This file should contain **only** what the compiler needs for generated code (the builtin descriptors that it doesn't generate itself, and definitions of things like `JubjubPoint` that are not defined as Compact types that the compiler can generate definitions for).

(I know there are examples of things like this in this file.  They should not be here.)

runtime/src/compact-types.ts
```diff
@@ -305,9 +315,9 @@ export const CompactTypeSecp256k1Base: CompactType<bigint> = {
     if (low192 >= 6277101735386680763835789423207666416102355444464034512896) {
       throw new CompactError('expected Secp256k1Base');
     }
-    let res = high64 << 192n | low192;
+    let res = (high64 << 192n) | low192;
```
kmillikin: This change and the one below it are inconsistent.  This one *adds* parentheses that are not needed due to the precedence rules, and the other *removes* parentheses that are not needed due to the precedence rules.

You should either (1) have the parentheses for clarity in both places, (2) leave them off in both places, or (3) just leave this code alone so that we don't create spurious diffs (making `git blame` harder to track).

runtime/src/compact-types.ts
```diff
@@ -319,10 +329,9 @@ export const CompactTypeSecp256k1Base: CompactType<bigint> = {
       throw new CompactError('expected Secp256k1Base');
     }
     // The ZKIR representation subtracts 1 from the value.
-    value = (value == 0n) ? MAX_SECP256K1_BASE : value - 1n;
+    value = value == 0n ? MAX_SECP256K1_BASE : value - 1n;
     const mask = (1n << 192n) - 1n;
-    return ocrt.bigIntToValue(value & ((1n << 192n) -1n))
-      .concat(ocrt.bigIntToValue(value >> 192n));
+    return ocrt.bigIntToValue(value & ((1n << 192n) - 1n)).concat(ocrt.bigIntToValue(value >> 192n));
```
kmillikin: No, please revert this change.  I liked the indentation that I intentionally wrote because it makes it clear that we are putting two bigints together and makes it easy to read that this matches the alignment (which it must do).

runtime/src/compact-types.ts
```diff
@@ -362,10 +371,9 @@ export const CompactTypeSecp256k1Scalar: CompactType<bigint> = {
       throw new CompactError('expected Secp256k1Scalar');
     }
     // The ZKIR representation subtracts 1 from the value.
-    value = (value == 0n) ? MAX_SECP256K1_SCALAR : value - 1n;
+    value = value == 0n ? MAX_SECP256K1_SCALAR : value - 1n;
     const mask = (1n << 192n) - 1n;
-    return ocrt.bigIntToValue(value & mask)
-      .concat(ocrt.bigIntToValue(value >> 192n));
+    return ocrt.bigIntToValue(value & mask).concat(ocrt.bigIntToValue(value >> 192n));
```
kmillikin: Revert please.

kmillikin:
This needs a changelog entry, and it needs a version bump for the Compact runtime.

runtime/src/compact-types.ts
golddydev: reverted back. I went with 3 :)

runtime/src/compact-types.ts
kmillikin: I also wonder why we define `mask` and then don't use it here.  It seems unnecessary since it has only a single use.  I'd remove it and then make the code for the other field type below look the same as here (remove the constant and inline its definition at the use site).

runtime/src/compact-types.ts
golddydev: reverted 😅

runtime/src/compact-types.ts
golddydev: [dd6d423](https://github.com/LFDT-Minokawa/compact/pull/630/commits/dd6d423a8b771bb506b2b1787ec99615e7bd7795)

golddydev: Oh, let me use inline for both base and field `toValue` functions

golddydev: [cbaf76e](https://github.com/LFDT-Minokawa/compact/pull/630/commits/cbaf76e3ac5783758a45b156f1e00fabf1177451)

Removed mask and apply inline definition

compiler/zkir-v3-library.compact
golddydev: https://github.com/golddydev/compact/blob/0b4381e7068ae823862e814bbd8e81cf861279e9/compiler/zkir-v3-library.compact#L80

comments updated

runtime/src/built-ins.ts
golddydev: [0d79e5a](https://github.com/LFDT-Minokawa/compact/pull/630/commits/0d79e5adfaeb41ba2b63d4433705d8258a6a2d7e)

runtime/src/built-ins.ts
golddydev: Removed

runtime/src/built-ins.ts
golddydev: Removed `ecrecover`

runtime/src/compact-types.ts
golddydev: [4749022](https://github.com/LFDT-Minokawa/compact/pull/630/commits/47490222ba55e634a2611f5086a3d0578aa7d5e5)

Removed this type.

compiler/zkir-v3-library.compact
golddydev: [1cb6247](https://github.com/LFDT-Minokawa/compact/pull/630/commits/1cb6247bcea5c8672f62251e462011694d81fafc)

Added document for secp point and signature and verify

compiler/test.ss
golddydev: Should we accept `signingNonce`, instead of `recoveryId`? Imo, users normally know only recoveryId (e.g. ethereum's uncompressed siganture)

doc/api/CompactStandardLibrary/exports.md
```diff
@@ -582,6 +609,58 @@ Asserts that the signature is valid; fails if the signature does not verify.
...
+circuit secp256k1PointY(pt: Secp256k1Point): Secp256k1Base;
+```
+
+### `secp256k1EcdsaVerify`
+
+Verifies an ECDSA signature over the foreign secp256k1 curve. Takes a 32-byte
+message hash, a [`Secp256k1EcdsaSignature`](#secp256k1ecdsasignature), and a
+public key (a [`Secp256k1Point`](#secp256k1point)). Returns true if the
+signature is valid; false if the signature does not verify.
+
```
iquerejeta:
```suggestion

This circuit takes `msgHash` as given and does not constrain it to any message.
The verifier is expected to additionally constrain that value so it is bound to
the message actually being verified (e.g. by hashing the message in-circuit). 

```

iquerejeta (approved):

doc/api/CompactStandardLibrary/exports.md
```diff
@@ -582,6 +609,58 @@ Asserts that the signature is valid; fails if the signature does not verify.
...
+the signer's identity must additionally bind the key to one it already trusts,
+for example by comparing `secp256k1EthereumAddress(pk)` against a stored
+address.
+
+Both low-s and high-s signatures are accepted, as in
+[FIPS 186-5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) section
+6.4.2, which constrains `s` only to the interval `[1, n - 1]`. Consequently
+`(r, s)` and `(r, n - s)` both verify against the same public key. A circuit
+that treats a signature as a unique identifier, such as a replay guard, must
+constrain `s <= n / 2` itself.
```
iquerejeta: We should not add that. a signature cannot be used as a unique identifier. The signer can produce different signatures for the same message.

```suggestion
6.4.2, which constrains `s` only to the interval `[1, n - 1]`. Consequently
`(r, s)` and `(r, n - s)` both verify against the same public key.
```

runtime/src/built-ins.ts
golddydev: [05e7279](https://github.com/LFDT-Minokawa/compact/pull/630/commits/05e7279806983665701b225e8fc4dbc2c755f310)

Refactored test case to use recover builtin as witness

kmillikin (approved):

PR #632, opened 2026-07-22
Break pass files up into individual files, one per pass

dybvig:
Addresses issue 631


compiler/analysis-passes/check-sealed-fields.ss
```diff
@@ -0,0 +1,135 @@
+;;; This file is part of Compact.
+;;; Copyright (C) 2025 Midnight Foundation
```
kmillikin, line 2: I don't think this can be Midnight Foundation, I think they transferred everything to LFDT.  We could probably change it as a separate change.  As far as I know the most recent LFDT guidance is here: https://lf-hyperledger.atlassian.net/wiki/spaces/TSC/pages/21430874/Copyright+and+License+Policy which is stale (Hyperledger became LFDT).

We should probably adopt something like "Copyright the Minokawa contributors".

I try to put the current year on new files, but this isn't exactly a new file.

kmillikin (approved):
I spot checked a few things but the review is mostly a rubber stamp, assuming that these are all just moved.

PR #649, opened 2026-07-23
[Issue 648] Forward compact dev tool to compactc for `compact compile --help`

### LINKED ISSUE #648: Expose `--feature-zkir-v3` in compact developer tool help
`compact compile --help` doesn't expose the `--feature-zkir-v3` flag and it should.

pataei:
Closes #648 .


kmillikin (approved):
As far as I understand it, this will kick off a 0.5.2 bug fix release.  On the one hand, this is a bug fix (to my mind).

On the other hand, it is a breaking API change.  There could be a script that sniffs the help text (for some reason) and which is now broken.

On another hand, scripts like that will be broken in the exact same way without any devtools version update if one of the non-builtin subcommands changes its help text.

I think we have three options here: (1) release a bug fix 0.5.2, (2) release a breaking 0.6.0 (with only this new feature, I guess, but that's OK), (3) hold off on releasing this until some later time.

I sort of prefer (2) (version numbers are so cheap that they're practically free), but I don't have a strong preference.  We should just be deliberate.

pataei:
I think there should be a separation between development and release, that is, the development test should not rely on a release and bump in version should not require an immediate release. For now, I'll go ahead and merge this but hold off on a release until a later time.

PR #650, opened 2026-07-23
Enable binary arithmetic for secp256k1 field types

kmillikin:
The binary operators `+`, `-`, and `*` now work for the types `Secp256k1Base` and `Secp256k1Scalar`.  They take operands of the same type and produce a result of that type.

The standard library circuits `add` and `mul` are removed.


compiler/analysis-passes/infer-types.ss
```diff
@@ -656,11 +657,15 @@
                      [(tunsigned ,src2 ,nat)
                       (make-native-field-op)]
                      [else (invalid-combination)])]
+                  [(tfield ,src1 (field-base (curve-secp256k1)))
+                   (nanopass-case (Ltypes Type) unaliased-type2
+                     [(tfield ,src2 (field-base (curve-secp256k1)))
+                      (values (k unaliased-type1 expr1 expr2) unaliased-type1)]
```
dybvig: Why is this the unaliased type instead of the original type?  It seems like this makes the `type` field unusable for an optimization pass that folds arithmetic operators.

dybvig (changes requested):

dybvig:
I wish I had communicated better here. I started Parisa on two-step process to simplify the switch from mbits/bits to type.  The first step is in this PR: https://github.com/LFDT-Minokawa/compact/pull/618, and simply removes mbits/bits from the pretty-printed versions of the language and thus in tests.ss.  The second was going to be to replace mbits/bits by the full type.  If you had started from that, there would be many fewer diffs in tests.ss to go over here.  But now that you've fought through the tests.ss updates with type present, it's probably better to keep it there.

kmillikin:
> I wish I had communicated better here. I started Parisa on two-step process to simplify the switch from mbits/bits to type. The first step is in this PR: #618, and simply removes mbits/bits from the pretty-printed versions of the language and thus in tests.ss. The second was going to be to replace mbits/bits by the full type. If you had started from that, there would be many fewer diffs in tests.ss to go over here. But now that you've fought through the tests.ss updates with type present, it's probably better to keep it there.

Ahhh.  That feels like two updates to the tests (1) rebase everything to remove mbits, (2) rebase everything to add a type.

kmillikin:
It's also easier this way to spot check the expectations and see that `mbits=n` has been correctly replaced with `tunsigned=m` and `#f` has been replaced by `tfield=field-native`.

dybvig:
> Ahhh. That feels like two updates to the tests (1) rebase everything to remove mbits, (2) rebase everything to add a type.

No, because our plan was to leave the type out permanently, like we do with some of the other cached-information fields.  But since you've got the type in, we might as well keep it.  There's always a tradeoff with including more of the IR information in the output.  It can allow some bugs to be caught earlier, but it can also require more rebasing of the expected output, which in turns makes PRs harder to evaluate.  It might or might not help one to understand the output of the pass.  More information can be useful but can also be overwhelming.  I am tending toward including less information these days except when actively debugging.

dybvig:
> It's also easier this way to spot check the expectations and see that mbits=n has been correctly replaced with tunsigned=m and #f has been replaced by tfield=field-native.

Yup.

compiler/analysis-passes/infer-types.ss
kmillikin: The types wouldn't work out very easily, because we will have inserted casts.  For example:

```compact
new type Base = Secp256k1Base;
export circuit test(b0: Base, b1: Base, b2: Base): Base {
  return b0 + b1 * b2;
}
```

after `infer-types`:

```scheme
(safe-cast (talias #t Base
             (tfield (field-base (curve-secp256k1))))
           (tfield (field-base (curve-secp256k1)))
 (+ (tfield (field-base (curve-secp256k1)))
     %b0.1
     (safe-cast (talias #t Base
                  (tfield (field-base (curve-secp256k1))))
                (tfield (field-base (curve-secp256k1)))
       (* (tfield (field-base (curve-secp256k1)))
          %b1.2
          %b2.3))))
```

I did it this way mostly because that's how `mbits` worked, but it made sense to me because we are actually performing arithmetic on a concrete type.

The type is used in this PR for instruction selection during `print-typescript`, so there we do want to see the operation type instead of having to follow a chain of aliases.

kmillikin: (Though actually, looking at that, it doesn't seem quite right.)

dybvig: Actually, I believe the aliases matter only for types that appear in the generated TypeScript definition file, and these do not.

compiler/analysis-passes/check-types-Lnodca.ss
```diff
@@ -126,7 +126,7 @@
          (nanopass-case (Lnodca Field-Type) ftype2
            [(field-scalar ,ctype2) (same-curve-type? ctype1 ctype2)]
            [else #f])]))
-    (define (sametype? type1 type2)
+    (define (same-type? type1 type2)
```
dybvig, line 129: In the future, I would prefer for the sake of reviewing that cosmetic changes, e.g., unnecessary renaming and reformatting, be made in a separate PR when the PR already has a lot of other changes.

Just FYI, I chose `sametype?` by analogy to `subtype?' but don't actually care.

compiler/analysis-passes/check-types-Lnodca.ss
```diff
@@ -560,27 +557,27 @@
...
        (with-output-language (Lnodca Type)
          `(tvector ,src ,len ,elt-type)))]
-    [(+ ,src ,mbits ,expr1 ,expr2)
-     (arithmetic-binop src "+" mbits expr1 expr2)]
-    [(- ,src ,mbits ,expr1 ,expr2)
-     (arithmetic-binop src "-" mbits expr1 expr2)]
-    [(* ,src ,mbits ,expr1 ,expr2)
-     (arithmetic-binop src "*" mbits expr1 expr2)]
+    [(+ ,src ,type ,expr1 ,expr2)
+     (arithmetic-binop src '+ type expr1 expr2)]
```
dybvig, line 576: FYI, I used a string here instead of a symbol because `arithmetic-biinop` doesn't dispatch on the operator name `op` in this pass as it does in `infer-types`.

compiler/circuit-passes/optimize-circuit.ss
```diff
@@ -164,9 +164,14 @@
       (define (nontriv-single-hash test.single)
         (triv-hash (car test.single)
           (nanopass-case (Lflattened Single) (cdr test.single)
-            [(+ ,mbits ,triv1 ,triv2) (mbits-hash mbits (commutative-triv-hash triv1 triv2 119001092))]
-            [(- ,mbits ,triv1 ,triv2) (mbits-hash mbits (triv-hash triv1 (triv-hash triv2 410225874)))]
-            [(* ,mbits ,triv1 ,triv2) (mbits-hash mbits (commutative-triv-hash triv1 triv2 513566316))]
+            [(+ ,primitive-type ,triv1 ,triv2)
+             ;; TODO(kmillikin): the result type is not determined by the types of the inputs, so we
```
dybvig: It's good to leave some sort of note here, but I think it's okay not to include primitive-type in the hash.  Post optimization, it's probably impossible for the result types to be different if the nontrivial operands are the same, and even if it is possible, there's no harm done (except for possibly slowing hashtable lookups) if a hash function returns the same value for two nonequal items.

compiler/circuit-passes/check-types-Lflattened.ss
```diff
@@ -363,12 +380,12 @@
...
-     (arithmetic-binop "-" mbits triv1 triv2)]
-    [(* ,mbits ,triv1 ,triv2)
-     (arithmetic-binop "*" mbits triv1 triv2)]
+    [(+ ,primitive-type ,triv1 ,triv2)
+     (arithmetic-binop '+ primitive-type triv1 triv2)]
+    [(- ,primitive-type ,triv1 ,triv2)
+     (arithmetic-binop '- primitive-type triv1 triv2)]
+    [(* ,primitive-type ,triv1 ,triv2)
+     (arithmetic-binop '* primitive-type triv1 triv2)]
     [(< ,bits ,triv1 ,triv2)
```
dybvig, line 389: I was expecting `bits` would also be replaced by `type`.  But since you don't need that, it should be done in a separate PR.

compiler/circuit-passes/resolve-indices-simplify.ss
```diff
@@ -452,40 +452,52 @@
...
-         (if mbits
-             a ; guaranteed by infer-types not to overflow
-             (modulo a (+ (max-field) 1)))))
+         (nanopass-case (Lnovectorref Type) type
+           ;; infer-types should guarantee that these are the only possibilities.
+           [(tfield ,src (field-native)) (modulo a (1+ (max-field)))]
+           [(tfield ,src (field-base (curve-secp256k1))) (modulo a (1+ (max-secp256k1-base)))]
+           [(tfield ,src (field-scalar (curve-secp256k1))) (modulo a (1+ (max-secp256k1-scalar)))]
+           [(tunsigned ,src ,nat) a]  ; Guaranteed by infer-types not to overflow.
+           [else (assert cannot-happen)])))
```
dybvig: So instead of the comment above, you could replace `(assert cannot-happen)` with `(asserf cannot-happen "infer-types should guarantee that these are the only possibilities") or something more informative.

compiler/analysis-passes/check-types-Lnodca.ss
kmillikin: Yeah, I figured that.  But I found the different API's for to be confusing.  At one point I assumed I was getting passed a symbol into a helper and used `eq?` and it took me too long to figure out what wasn't working.

compiler/analysis-passes/infer-types.ss
kmillikin: Are you suggesting that for non-exported nominal aliases, we can type check them as nominal aliases but generate code from `infer-types` that treats them as if they were structural aliases?

For now we need the casts to ensure that the types work out.  To me it's most direct to say that arithmetic is defined only for specific builtin types and that arithmetic for nominal aliases works via casting.  The other view is that a nominal type alias introduces overloads for the arithmetic operators available for the underlying type, and we could certainly implement it that way.

I don't think there's a difference for constant folding operations except whether you have to look through a chain of operand casts to see if the operands are constant or else you have to look through a chain of operation type casts of some kind to figure out which underlying arithmetic to use.

compiler/circuit-passes/optimize-circuit.ss
kmillikin: I'll just change it to a note then.

compiler/circuit-passes/check-types-Lflattened.ss
kmillikin: Agreed.

compiler/circuit-passes/resolve-indices-simplify.ss
kmillikin: Done and also at the other sites.  The message identifies the failure site.

compiler/zkir-v3-passes/reduce-to-zkir.ss
```diff
@@ -1007,17 +1007,17 @@
       (assert cannot-happen)])
 
   (Single : Single (ir var-name instr*) -> * (instr*)
-    [(+ ,mbits ,triv0 ,triv1)
+    [(+ ,primitive-type ,triv0 ,triv1)
```
dybvig, line 1010: I feel like I should know this already, but why aren't any zkir changes required to support the additional operations?

runtime/src/built-ins.ts
```diff
@@ -51,7 +51,7 @@ export function subField(x: bigint, y: bigint): bigint {
   // (x - y) % FIELD_MODULUS would return an incorrect value for negative values of x - y.
   // also, any implementation involving % would likely be more expensive
   const t = x - y;
-  return t >= 0 ? t : t + FIELD_MODULUS;
+  return t >= 0n ? t : t + FIELD_MODULUS;
```
dybvig, line 54: I guess this doesn't matter, but I like the change anyway.

dybvig (approved):
Approved with just a question about why (nontrivial) zkir changes aren't required.

compiler/zkir-v3-passes/reduce-to-zkir.ss
kmillikin: We decided to make operations (ad hoc) polymorphic in ZKIR v3.  Instructions can have a set of types that they work for and the implementation will be chosen based on the type.

For example, you can see here all the types that `add` works for:

https://github.com/midnightntwrk/midnight-ledger/blob/04c9c5d9bcebb8d4427d8589fb54d58a55599c14/zkir-v3/src/ir.rs#L752-L767

PR #651, opened 2026-07-23
[issue-117] rejects all effect producing operations in the constructor

### LINKED ISSUE #117: Cannot mint shielded tokens in contract constructor
### Context & versions:

Compact: 0.29
Node toolkit: 0.21.0-0b68aa0e

Following circuits in contract (shielded):

```

export ledger mint_count: Counter;

constructor() {
    const domain = pad(32, "bug:shielded:ctor");
    mintShieldedToken(
        domain,
        100 as Uint<64>,
        evolveNonce(0 as Uint<128>, domain),
        right<ZswapCoinPublicKey, ContractAddress>(kernel.self())
    );
    mint_count.increment(1 as Uint<16>);
}

export circuit get_count(): Uint<64> {
    return mint_count.read();
}

and unshielded:

export ledger mint_count: Counter;

constructor() {
    const domain = pad(32, "bug:unshielded:ctor");
    mintUnshieldedToken(
        domain,
        100 as Uint<64>,
        left<ContractAddress, UserAddress>(kernel.self())
    );
    mint_count.increment(1 as Uint<16>);
}

export circuit get_count(): Uint<64> {
    return mint_count.read();
}
```

### Steps to reproduce:

Example is here: [https://github.com/midnightntwrk/midnight-contracts/tree/added_toolkit](url)

1. Deploy contract
2. Execute get_count circuit

### Actual behavior:

Unshielded mint works, while shielded fails with:

`Error: DynamicError { error: FailedProvingTx(EffectsCheckFailure(CommitmentsNEClaimedShieldedReceives { commitments: [(0, Commitment(9a63908e4f2c23618089c871c23bfc0cdf729ec50678fb1e42df156cd46aedb5), ContractAddress(0000000000000000000000000000000000000000000000000000000000000000))], claimed_shielded_receives: [] })) }`

### Expected behavior:

Either both pass or both fail

pataei:
Closes #117. See changelog.


compiler/analysis-passes/reject-constructor-effects.ss
```diff
@@ -15,35 +15,58 @@
...
-      (function-name emit-condition-function-name)
-      (src emit-condition-src)
-      (reason emit-condition-reason))
+    (define-condition-type &effect-condition &condition
+      make-effect-condition effect-condition?
+      (function-name effect-condition-function-name)
+      (src effect-condition-src)
+      (reason effect-condition-reason))
+    ; Kernel update operations that emit a transaction-level effect
+    (define forbidden-kernel-ops
```
dybvig, line 28: I wonder if it makes sense to make any kernel calls.  Maybe they should all be forbidden.  If not, the `function` form in the ledger ADT declarations should include a flag that says whether calls by the constructor to the function are allowed.

compiler/analysis-passes/reject-constructor-effects.ss
```diff
@@ -15,35 +15,58 @@
...
+        claimZswapCoinSpend
+        claimZswapNullifier
+        claimUnshieldedCoinSpend
+        incUnshieldedInputs
+        incUnshieldedOutputs
+        claimContractCall
+        checkpoint))
+    ; native entries that create Zswap coin inputs/outputs
+    (define forbidden-native-functions
+      '("__compactRuntime.createZswapInput"
```
dybvig, line 42: Does it make sense to call `ownPublicKey`?  If not, maybe any native marked as a witness should be forbidden.  Otherwise, the `define-native` form should have an additional field that says whether calls to the native are allowed in the constructor.

compiler/analysis-passes/reject-constructor-effects.ss
```diff
@@ -68,30 +94,38 @@
...
-         (when (emit-condition? result)
-           (let ([offending-function-name (emit-condition-function-name result)])
+         (when (effect-condition? result)
+           (let ([offending-function-name (effect-condition-function-name result)])
              (if (eq? offending-function-name #f)
-                 (source-errorf src "constructor cannot emit an event but ~a at ~a"
-                                (emit-condition-reason result)
-                                (format-source-object (emit-condition-src result)))
-                 (source-errorf src "constructor cannot emit an event but calls (directly or indirectly) ~a, which ~a at ~a"
+                 (source-errorf src "constructor cannot call an effect-producing operation but ~a at ~a"
```
dybvig, line 100: `kernel.self`, at least, is not effect-producing.  We need to come up with a more generic error message or simplify the error message to say that the "constructor cannot call ~a but does so at ~a"  where the first "~a" is the name of the forbidden native or kernel operation.

compiler/analysis-passes/reject-constructor-effects.ss
```diff
@@ -68,30 +94,38 @@
...
+                                (effect-condition-reason result)
+                                (format-source-object (effect-condition-src result)))
+                 (source-errorf src "constructor cannot call an effect-producing operation but calls (directly or indirectly) ~a, which ~a at ~a"
```
dybvig, line 103: Ditto

compiler/analysis-passes/reject-constructor-effects.ss
```diff
@@ -15,35 +15,58 @@
...
       (let ([a (eq-hashtable-cell function-ht function-name #f)])
+        (when (eq? (cdr a) 'forbidden-native)
+            (raise (make-effect-condition function-name src (format "calls ~a" (id-sym function-name)))))
         (process-circuit! a)
         (let ([result (cdr a)])
           (assert (not (eq? result 'inprocess-circuit)))
-          (when (emit-condition? result)
-            (raise-continuable result)))))
+          (when (effect-condition? result)
+            (raise-continuable (make-effect-condition function-name
```
dybvig, line 67: This won't work properly when the constructor calls `foo`, `foo` calls `bar`, and `bar` calls a forbidden function or operator.  The error message will always identify the function called directly from the constructor, so in this case, it will say the constructor calls (directly or indirectly) `foo`, which calls the forbidden operator at some source position within the definition of `bar`.

compiler/analysis-passes/reject-constructor-effects.ss
```diff
@@ -15,35 +15,58 @@
...
+          (guard (c [(effect-condition? c) (set-cdr! a c)]
                     [else (raise-continuable c)])
             (set-cdr! a 'inprocess-circuit)
             (Expression maybe-expr function-name)
             (set-cdr! a #f)))))
-    (define (process-function-name! function-name)
+    (define (process-function-name! function-name src)
       (let ([a (eq-hashtable-cell function-ht function-name #f)])
+        (when (eq? (cdr a) 'forbidden-native)
+            (raise (make-effect-condition function-name src (format "calls ~a" (id-sym function-name)))))
```
dybvig, line 62: When the constructor calls a forbidden native directly, this will lead to error messages that say that constructor calls (directly or indirectly) <forbidden-native>, which calls <forbidden-native> at some character-position within the constructor.

I see an instance of this in test.ss at line 31079.

dybvig, line 62: Also, this line is indented improperly.

compiler/analysis-passes/reject-constructor-effects.ss
```diff
@@ -15,35 +15,58 @@
 
 #!chezscheme
 
-(define-pass reject-constructor-emit : Lnodca (ir) -> Lnodca ()
-  ; this pass raises an exception if the constructor attempts an emit
+(define-pass reject-constructor-effects : Lnodca (ir) -> Lnodca ()
```
dybvig, line 18: This pass seems like a good candidate to use `pending-errorf` instead of `source-errorf`.

dybvig (changes requested):

compiler/analysis-passes/reject-constructor-effects.ss
pataei: Checked with Andrzej:
"ownPublicKey is a "regular" witness, just integrated into runtime, so it's fine in the constructor. For kernel operations - I would need to check in detail, but many kernel operations assume either already initialized contract or emit ledger effects."

PR #653, opened 2026-07-24
Allow `Uint<0>`

kmillikin:
The type `Uint<0>` was forbidden both in the language reference and in the compiler implementation.  But according to the rule for sized unsigned integer types it would be equivalent to the bounded unsigned integer type `Uint<0..1>` which is allowed.  Therefore, there is no real reason to disallow it.


dybvig (approved):
LGTM.

PR #662, opened 2026-07-28
Make curve points into builtin types

kmillikin:
Previously the curve point types `JubjubPoint` and `Secp256k1Point` were nominal (opaque) type aliases for builtin Opaque types.  This was weird (that they were aliases) and the implementation was error prone because they could not use the normal handling of JS opaque types.

Instead, implement them as real builtin Compact types.  The fixup support for renaming `NativePoint` to `JubjubPoint` was removed along with all its testing because (1) that change was several revisions ago and (2) fixup can't currently rename a standard library struct to a builtin type.


compiler/parser.ss
```diff
@@ -633,6 +635,10 @@
        (lambda (src kwd)
          (with-output-language (Lparser Type)
            `(tfield ,src ,kwd)))]
+      [type-jubjub-point :: src (KEYWORD JubjubPoint) =>
```
dybvig: I think we should not go down the path of adding these as reserved words.  It should be straightforward to inject definitions of new type names into the standard library in a way analogous to how natives are injected.  That is, we have a midnight-native-types.ss that produces a set of intermediate-language definitions that look something like `(define-native-type name type)`, e.g., `(define-native-type JubJubPoint (tpoint <stdlib-src> (curve-jubjub)))`.  We can do the same for scalar types.

dybvig (changes requested):

compiler/parser.ss
dybvig: Maybe `midnight-inlines.ss` is a better example to follow.

kmillikin:
Moving these out of the language and into the standard library is a clear improvement.  I think this is ready for re-review.

dybvig (approved):

Issue #638, opened 2026-07-22
Export (some) descriptors from a compiled contract


kmillikin:
We use descriptors as a runtime representation of a Compact type, they are objects with methods for converting between JS representations of Compact values and ledger FAB (field-aligned binary) representations.

Some of the exported Compact runtime functions (usually ones that wrap an on-chain runtime version that requires a FAB input) need to be passed a descriptor, for example `persistentHash` for SHA-256 hashing.

These are difficult for a DApp to work with because it either needs to define its own descriptors that are equivalent to the compiler-generated ones, or else it needs to modify the compiler-generated code to export (and possibly rename) the cryptically named ones in the compiled contract.

We should come up with a way, for exported types, to also export the corresponding descriptor.

PR #663, opened 2026-07-29
Restructure the Compact runtime libraries

kmillikin:
- `compact-types.ts` holds types emitted by the compiler (like `JubjubPoint` and `Secp256k1Point`) and the base built-in descriptors used by the compiler to build all emitted descriptors.  The `toBinaryRepr` helper stays there because it has to do with FAB.  Everything else is either deleted because it is unused (and duplicates the compiler generated descriptors) or is moved elsewhere if it's needed for some other part of the runtime.

- `built-ins.ts` holds functions emitted by the compiler in `print-typescript` and to implement the Midnight natives.  Utility-style functions have been moved to `utils.ts`.

This is a breaking change because some unnecessary and unused exported types and descriptors are removed, and some exported helpers are no longer exported because they're moved to where they're needed.


kmillikin:
I moved or deleted the code, but otherwise did not change it.

dybvig (approved):
LGTM

JosephDenman (commented):
I see the benefit of removing unused descriptors. I do not see the benefit of moving the remaining descriptors from `compact-types.rs` into various other files in the other files in the runtime package. With all/most of the descriptors in `compact-types.ts`, a developer can easily see which descriptors have already been defined when they're writing a new utility function, reducing the chances they accidentally implement the same descriptor twice.

kmillikin:
> I see the benefit of removing unused descriptors. I do not see the benefit of moving the remaining descriptors from `compact-types.rs` into various other files in the other files in the runtime package. With all/most of the descriptors in `compact-types.ts`, a developer can easily see which descriptors have already been defined when they're writing a new utility function, reducing the chances they accidentally implement the same descriptor twice.

Yeah, I can see that view.  But these aren't used by the compiler and they aren't generated by the compiler --- there's no guarantee that they actually match the implementation in the standard library except that we've supposedly tested them.

Ideally we'd find some way to generate these (part of what https://github.com/LFDT-Minokawa/compact/issues/638 is about) and DApp developers would never have any reason to define their own descriptors.

I thought I'd move them now because I'd done the work of garbage collection and we would know that these were ones we could eventually eliminate.  But I can move them back if you wish.

kmillikin:
I'll add that I would expect developers to look at the docs to see which descriptors were available, not the source code (they'd have to be quite savvy to know to look in `runtime/src/compact-types.ts` for them).

PR #669, opened 2026-07-31
Implement JS runtime type checks on JS opaque values

kmillikin:
The generated JS code now has circuit argument and witness return value type checks for the JS opaque types `Opaque<'string'>` and `Opaque<'Uint8Array'>`.  Before, we allowed any value at all to be passed or returned.  This is a **breaking change** for programs that relied on being able to store any random JS value as, say, an `Opaque<'string'>`.


JosephDenman:
Please add a PR description describing why these changes are needed / what they're trying to accomplish.

compiler/test.ss
```diff
@@ -90342,7 +90343,25 @@ groups than for single tests.
...
+      "export circuit test1(srt: Opaque<'string'>): [] { return; }"
+      )
+    (stage-javascript
+      '(
+        "test('argument type checks on JS opaque types', async () => {"
+        "  const [contract, context] = await startContract(contractCode, {}, 0);"
+        "  expect((await contract.circuits.test0(context, new Uint8Array(32))).result).toEqual([]);"
+        "  await expect(contract.circuits.test0(context, 0n)).rejects.toThrow();"
+        "  expect((await contract.circuits.test1(context, 'Hello, Compact!')).result).toEqual([]);"
+        "  await expect(contract.circuits.test1(context, 0n)).rejects.toThrow();"
```
JosephDenman: Please add some text to the `toThrow` calls to check that the correct error is thrown. We don't follow this discipline everywhere in `test.ss`, but new tests should. It doesn't have to be the exact error message string, but enough to identify the error.

compiler/typescript-passes/print-typescript.ss
```diff
@@ -1113,7 +1113,14 @@
                        var field-name))]
                   [(tunsigned ,src ,nat) (format "typeof(~a) === 'bigint' && ~:*~a >= 0n && ~:*~a <= ~dn" var nat)]
                   [(tbytes ,src ,len) (format "~a.buffer instanceof ArrayBuffer && ~:*~a.BYTES_PER_ELEMENT === 1 && ~:*~a.length === ~s" var len)]
-                  [(topaque ,src ,opaque-type) "true"]
+                  [(topaque ,src ,opaque-type)
+                   (case opaque-type
+                     [("string") (format "typeof (~a) === 'string'" var)]
+                     [("Uint8Array") (format "~a instanceof Uint8Array" var)]
+                     [("JubjubPoint" "Secp256k1Point") "true"]
```
JosephDenman, line 1126: Is it correct to default to `true` for `JubjubPoint` and `Secp256k1Point`? We don't want _some_ check that `var` is actually one of those?

compiler/typescript-passes/print-typescript.ss
```diff
@@ -1113,7 +1113,14 @@
...
+                     [else (assertf cannot-happen
```
JosephDenman, line 1127: Maybe we want a test in the E2E tests that confirms this final branch case works as expected?

JosephDenman (changes requested):

compiler/typescript-passes/print-typescript.ss
kmillikin: It's not correct but it's changed in #662.

compiler/typescript-passes/print-typescript.ss
kmillikin: We can't really write a test to activate unreachable code (unless it were actually reachable....)

compiler/test.ss
kmillikin: Done.

JosephDenman (approved):
LGTM

PR #670, opened 2026-07-31
Change `Opaque<'Uint8Array'>` JS equality to be more in line with ZKIR

kmillikin:
In ZKIR we use the Poseidon hash of the (ledger FAB encoding of the) contents, but in JS we used strict equality.  For typed array, this is object reference equality.

Changed here to be (1) same length and (2) element-wise strict equality.


compiler/test.ss
```diff
@@ -90342,7 +90342,35 @@ groups than for single tests.
         "});"
         ))
     )
-  )
+
+  (test
+    '(
+      "export circuit test(arr0: Opaque<'Uint8Array'>, arr1: Opaque<'Uint8Array'>): Boolean {"
```
JosephDenman: This is a pure circuit and `checkProofData` ZKIR checks only run on impure circuits. I think a more comprehensive test would touch a counter or something to get impurity.

doc/compact-reference.mdx
```diff
@@ -2554,9 +2558,13 @@ has a specific structure type, so does the other.
   the values are equal if the values of the corresponding fields are equal.
 - **Enumeration types:**
   the values are equal if they are the same enumeration member.
-- **`Opaque` types:**
+- **`Opaque<'string'>:**
```
JosephDenman: The opening backticks are never closed in this line and the opaque `Uint8Array` line below.

compiler/test.ss
JosephDenman: We should have a test for `!=` as well, and for nested `Uint8Array`, for example `Vector<2, Opaque<'Uint8Array'>>`.

compiler/test.ss
```diff
@@ -90342,7 +90342,35 @@ groups than for single tests.
...
+      "  return arr0 == arr1;"
+      "}"
+      )
+    (stage-javascript
+      '(
+        "test('Uint8Array equality comparison', async () => {"
```
JosephDenman, line 91129: We should also have a test case exercising empty `Uint8Array` equality.

JosephDenman: Even better, a test involving three circuits all impure:

```

ledger checkResult: Boolean;

export circuit arraysEqual(a: Opaque<'Uint8Array'>, b: Opaque<'Uint8Array'>): Boolean {
  const result = a == b;
  checkResult = disclose(result);
  return result;
}

export circuit arraysNotEqual(a: Opaque<'Uint8Array'>, b: Opaque<'Uint8Array'>): Boolean {
  const result = a != b;
  checkResult = disclose(result);
  return result;
}

export circuit equalsDefault(a: Opaque<'Uint8Array'>): Boolean {
   const result = a == default<Opaque<'Uint8Array'>>;
   checkResult = disclose(result);
   return result;
}
```

with a number of the combinatorial true / false cases and including a test case like:

```
expect((await C.circuits.equalsDefault(Ctxt, new Uint8Array(0))).result).toEqual(true);
```

The `default` case falls nicely out of this change because `default<Opaque<'Uint8Array'>>` lowers to `0` and `compress([]) == 0` but previously the above test would fail I think.

compiler/test.ss
```diff
@@ -90342,7 +90342,35 @@ groups than for single tests.
         "});"
         ))
     )
-  )
+
+  (test
+    '(
+      "export circuit test(arr0: Opaque<'Uint8Array'>, arr1: Opaque<'Uint8Array'>): Boolean {"
+      "  return arr0 == arr1;"
```
JosephDenman: We also probably want a test that writes to a ledger field and reads it back out and checks that equality still holds. Under the previous semantics, I believe an opaque value could never be equivalent to the same opaque value just written to and read from the ledger. We should add a regression for that case.

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
...
 
+## [Toolchain 0.33.112, language 0.25.104, runtime 0.18.102]
+
+### Changed
+
+- Equality of `Opaque<'Uint8Array'>` is now (1) same length and (2) element-wise
+  strict equality (`===`).  It was formerly simple strict equality, which for
+  typed arrays is object reference equality.
+
+  This is a breaking change in the language, because `Uint8Arrays` that were
```
JosephDenman, line 16: You say this is a breaking change, but only the bug fix component of the semver was updated. Doesn't this mean that a user with `pragma language_version 0.25` will silently pick up the new semantics?

JosephDenman (changes requested):

compiler/test.ss
kmillikin: Yeah, that would be a more comprehensive test.  The intent of this one is to check the semantics of the JS code.

doc/compact-reference.mdx
kmillikin: Thanks.

compiler/test.ss
kmillikin: != and nested types use the same code.  We don't have those tests for any other types.

compiler/test.ss
kmillikin: Empty Uint8Array equality trivially falls into the "same length" test, which is already tested.

CHANGELOG.md
kmillikin: Yeah, a user with that pragma who builds the compiler from the dev branch `main` will silently pick up the new semantics (and the same thing for compiler version and runtime version updates).  It's really a problem with the way we assign version numbers on `main` and not something that I'm going to redesign as part of this PR.

compiler/test.ss
JosephDenman: Fair on nesting. I'll drop that one.

`!=` isn't quite the same code though. The helper body is shared, but the decision of whether to call the helper at all is duplicated. There's one copy of the type predicate in the `==` clause and a separate copy in the `!=` clause (`print-typescript.ss` ~2737 and ~2761), and this PR edits both. So, a slip in the `!=` copy wouldn't be caught by anything.

And on not having these tests for other types. We do, for the two cases most like this one. The `JubjubPoint` test on line 72763 has both `pointsEqual` and `pointsNotEqual`, and the `Secp256k1Point` test on line 91001 has the same pair. Both were added for exactly this situation, right? An opaque type whose equality changed from reference to structural. Following that shape here seems more consistent than not.

compiler/test.ss
JosephDenman: Agreed on empty-vs-empty. So that's fine.

`equalsDefault` is a different thing though, and I don't think it's a length comparison at all. In the circuit, `a == default<Opaque<'Uint8Array'>>` is `compress(a) == 0` against the field literal. There's no length check involved. In JS it's the length plus element-wise path. Those two agree only because of the empty special case in `transient-crypto/src/fab.rs`:

```
AlignmentAtom::Compress => {
    // Special case for the empty string to make defaults work well.
    if self.0.is_empty() {
        writer.write(&[0.into()]);
```

The reason I think it's worth having is that it's a genuine regression test rather than extra coverage. Before this change, `default<Opaque<'Uint8Array'>>` was `new Uint8Array(0)` in JS and `x === new Uint8Array(0)` is false for any distinct object. So JS said false while the circuit said true. The test fails before the change and passes after, which is the divergence the PR is fixing, and I don't think anything else in the suite demonstrates that.

That's also why I suggested making the circuits impure with the ledger `checkResult` write. As written the new test's circuit is pure, so `checkProofData` never runs and the ZKIR side isn't exercised at all — same reason the Jubjub and Secp256k1 equality tests write to the ledger.

JosephDenman (approved):
LGTM

compiler/test.ss
kmillikin: Yeah, you're right about the duplication.  Seems like we should eliminate that for sure.

PR #672, opened 2026-07-31
It's an error to try to get coordinates from the secp256k1 identity p…

kmillikin:
…oint

Throw a `CompactError` in the Compact runtime when attempting to extract x- and y-coordinates from the secp256k1 identity point.  It doesn't have valid coordinates, and this behavior matches the ZKIR behavior.


JosephDenman (approved):
LGTM, just two comments.

One thing on the test: the test passes `{ x: 0n, y: 0n, identity: true }`, but the change log note is specifically about identity points that aren't canonicalized. Passing `{ x: 7n, y: 9n, identity: true }` would pin down that we're checking the flag rather than the coordinates, which is the part that could regress.

And one nit: `zkir-v3-natives.ss` lines 63 and 68 still say "the X coordinate of" / "the Y coordinate of". Those end up in disclosure messages.

PR #678, opened 2026-07-31
Update the ledger dependency to bring in ZKIR 3.1 work in progress

kmillikin:
We will need ZKIR 3.1 support for secp256r1 and Curve25519.

The `zkir-v3` and `zkir-v3-wasm` dependencies are changed to include the new features.  The `onchain-runtime-v4` dependency is kept to track the tag `ledger-9.1.0.0-rc.3`.

The toolchain version is bumped (if nothing else, it reports a different string for `--feature-zkir-v3 --ledger-version`).

`inputs.zkir.follows` is removed, because that input is removed upstream and the override will now trigger a nix warning.


JosephDenman (commented):
You said:

> The onchain-runtime-v4 dependency is changed to keep it in lockstep with zkir-v3 and zkir-v3-wasm (it could conceivably be left at 9.1.0.0-rc.3).

Then why not keep it at `9.1.0.0-rc.3`? You're saying the benefit of pinning it exactly is greater than the benefit of being able to know exactly which version of onchain runtime we're using without chasing hashes into another repo? If we're going to use hashes, then we should at least update the comment in the flake input `onchain-runtime-v4 = { ... }` to be accurate.

kmillikin:
> You said:
> 
> > The onchain-runtime-v4 dependency is changed to keep it in lockstep with zkir-v3 and zkir-v3-wasm (it could conceivably be left at 9.1.0.0-rc.3).
> 
> Then why not keep it at `9.1.0.0-rc.3`? You're saying the benefit of pinning it exactly is greater than the benefit of being able to know exactly which version of onchain runtime we're using without chasing hashes into another repo? If we're going to use hashes, then we should at least update the comment in the flake input `onchain-runtime-v4 = { ... }` to be accurate.

I'm happy to do that!  I've asked at least twice how we want to handle this and never gotten any answer!

These two comments always seemed kind of dubious to me, so I've always just ignored them :).  I'm not sure that the part about "flake update" is correct any more, I don't know why we care (here in our flake.nix) about the releases of the on-chain runtime for ledger 9.1.0.0-rc.2, saying "this is a tag" when you put a tag and "this is a commit hash" when you put a commit hash seems kind of pointless to me.  (Edit: the `zkir` comment says "dependency for compact-runtime release" which isn't even right.)

kmillikin:
> You're saying the benefit of pinning it exactly is greater than the benefit of being able to know exactly which version of onchain runtime we're using without chasing hashes into another repo?

I'm not saying anything like that!

I'm saying that on our dev branch, if we expect that there will be a release of on-chain runtime at or after this commit hash and before or at the same time as we release Compact toolchain 0.34 then we should start tracking code that will be in that on-chain runtime release.  We don't really need to try to debug mismatches between zkir-v3 `04c9c5d` and on-chain runtime `ledger-9.1.0.0-rc.3` if they occur.

Compact toolchain 0.34 will target ZKIR 3.1 (and it will likely have the ZKIR 2 backend removed).  So I'm suggesting that we should start moving the on-chain runtime dependency in lockstep with the ZKIR v3 key generation dependency (and the proof server release we will need).

But as I said, the alternative is that we leave the on-chain runtime behind the ZKIR 3.1 dependency and see what happens.

(Edit: and by the way: in both cases (tag and commit hash) the dependency is pinned exactly unless the ledger uses mutable tags, and in both cases you have to chase into the ledger repository to find out what version the on-chain runtime built from that tag or hash is claiming to be (the on-chain runtime version is not `ledger-9.1.0.0-rc.3`).  So that isn't really a consideration for me.)

JosephDenman (approved):
LGTM

Issue #679, opened 2026-08-03
Overload `+`, `-`, and `*` to work for secp256k1 field types


kmillikin:
As part of splitting up the work in the parent ticket, to review, commit, and QA test in stages, we should make the existing arithmetic operators work for the new secp256k1 field types.

kmillikin:
This was implemented in #650

mhounslow:
Looks good to me ✅ 

I added regression tests in [compact-end-2-end here](https://github.com/midnightntwrk/compact-end-2-end/pull/75).

The tests covered `+`, `-`, and `*` for both `Secp256k1Base` and `Secp256k1Scalar` which included:

- addition wraparound
- subtraction underflow
- multiplication wraparound. 

These calls asserted all three calculated modular results inside the proof are good, and the accepted state transitions were confirmed through indexed ledger counters.

PR #688, opened 2026-08-04
Test secp256k1 arithmetic operators

lowhung:
Adds compiler E2E coverage for #679.

Checks `+`, `-`, and `*` lowering for secp256k1 base and scalar fields.

Tests: lint and focused Vitest suite (1/1).

### DIFF
--- a/examples/secp256k1/arithmetic.compact
+++ b/examples/secp256k1/arithmetic.compact
@@ -1,5 +1,5 @@
 // This file is part of Compact.
-// Copyright (C) 2025 Midnight Foundation
+// Copyright (C) 2026 Midnight Foundation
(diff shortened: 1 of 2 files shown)


examples/secp256k1/arithmetic.compact
```diff
@@ -0,0 +1,47 @@
+// This file is part of Compact.
+// Copyright (C) 2025 Midnight Foundation
```
kmillikin: If you can remember, use 2026 ;)

kmillikin (approved):

examples/secp256k1/arithmetic.compact
mhounslow: Heh 😅 copy pasta I'll admit!

Issue #658, opened 2026-07-24
Cross-contract callees cannot perform shielded operations (currentZswapLocalState is `undefined`)


whankinsiv:
### Context & versions:

Contract-to-contract **shielded** value transfer. A caller contract sends a shielded coin to a recipient contract's address and in the same transaction calls the recipient so it can claim the receive. The recipient's claim runs as a cross-contract callee, and a callee is executed with no Zswap local state so any shielded operation inside it fails before a transaction is built. Confirmed as a bug / missing feature by @tkerber. 

The unshielded equivalent works, only the shielded path is affected because unshielded transfers never touch Zswap state.

Versions:
* compact-runtime: 0.18.0-rc.1
* compact-compiler: 0.33.0-rc.1
* ledger-9: 1.0.0-rc.3
* node: 2.0.0-rc.4

### Steps to reproduce:

`git clone https://github.com/midnightntwrk/compact-end-2-end.git`
`cd compact-end-2-end`
`git checkout origin/whankinsiv/shielded-contract-recipient`
`vp run cases:regression -- shielded-contract-recipient-send-atomic` (The case is pinned to a runtime error containing "Zswap local state is undefined for contract")

### Actual behavior:

The callee throws:
`Error: Zswap local state is undefined for contract '<B-address>'`
from `assertHasCurrentZswapLocalState` (`compact-runtime/dist/zswap.js`). No transaction is submitted.

The cause is in `setupCallContext` (`compact-runtime/dist/contract.js`), which blanks the callee's private and Zswap state before invoking it:

```
// Undefined because these two should only be called for sub-calls, which do not support witnesses
context.callContext.currentPrivateState = undefined;
context.callContext.currentZswapLocalState = undefined;
```

With `currentZswapLocalState` undefined, every shielded primitive in a callee fails the guard.

### Expected behavior:

A cross-contract callee should be able to claim a shielded receive so that a caller can send a shielded coin to a recipient contract and have the recipient claim it in the same transaction.

JosephDenman:
The compact side of this fix is done. We'll have to wait for the next release candidate to be published and Midnight.js and Midnight SDK to use the release candidate before the fixes filter through to dApp builders. I have branches with draft fixes in both of those repos.

mhounslow:
QAed and covered with a test in our compact-end-2-end repo [here](https://github.com/midnightntwrk/compact-end-2-end/pull/85).

pataei:
[...]

PR #694, opened 2026-08-05
Address issue #658

JosephDenman:
Cross-contract callees couldn't do anything with shielded coins. The runtime blanked a callee's Zswap local state before invoking it, so `receiveShielded`, `sendShielded` and related functions failed with "Zswap local state is undefined".

After this, with supporting Midnight.js and Midnight SDK changes, one contract can send a shielded coin to another, the recipient takes custody, re-spends part of it, and the value lands in the submitter's wallet. This is an E2E test I have verified with these changes.

Original GH issue: https://github.com/LFDT-Minokawa/compact/issues/658


CHANGELOG.md
```diff
@@ -5,6 +5,34 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Toolchain 0.33.118, language 0.25.107, runtime 0.18.106]
```
kmillikin: I'd bump the toolchain version (because the compiler will generate code that pulls in a new runtime version), then you don't need the "skip-changelog" label.

runtime/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@midnight-ntwrk/compact-runtime",
-  "version": "0.18.105",
+  "version": "0.18.106",
```
kmillikin, line 3: I'm sort of curious why we've been apparently bumping this one and not the one below, and if it's even necessary.  Something to sort out later.

runtime/test/cross-contract-zswap.test.ts
```diff
@@ -0,0 +1,144 @@
+// This file is part of Compact.
+// Copyright (C) 2025 Midnight Foundation
```
kmillikin: I try to remember to use 2026 here.  Maybe we should see if we can just remove the date.

kmillikin (approved):
LGTM with some nits mentioned.

runtime/package-lock.json
JosephDenman: I'm not sure. They shouldn't disagree AFAIK. I just regenerate the correct package lock with `npm install`.

CHANGELOG.md
JosephDenman: Done.

runtime/test/cross-contract-zswap.test.ts
JosephDenman: Fixed.

runtime/package-lock.json
kmillikin: Before your change, one was .105 and the other was .102 (in the same file).  I think the answer is that something cares about the first version matching `package.json` but that thing does not care about the second version --- so we've been updating it by hand to satisfy whatever that thing is, rather than regenerating :)

PR #696, opened 2026-08-05
[perf] Hash manifests in-process

mhounslow:
## Summary

- Hash manifest files in-process through OpenSSL’s EVP interface when `COMPACT_LIBCRYPTO` is configured.
- Configure Nix builds and development shells to use the pinned OpenSSL library.
- Retain the existing `sha256sum`/`shasum` fallback when no library is configured.

## Why

The manifest generation previously started a separate process for every file being hashed, the native implementation removes that per-file process cost, still preserving manifest generation, the `sha256-file` interface, and hash output.

## Testing

- `nix develop --command ./compiler/go`
- `nix build --no-link`
- Standard SHA-256 vectors, including the one-million-`a` vector
- Native and command-line fallback implementations
- Failure behavior for an invalid configured library


kmillikin:
I don't have the expertise to give this a very thorough review.  Redirecting to @dybvig.

compiler/test.ss
```diff
@@ -887,6 +887,32 @@ groups than for single tests.
                      (next-value (cdr v*))))))))]))
 )
 
+(module ()
```
dybvig, line 890: Rather than putting files in /tmp, I'd rather we just create files in a sha256-test subdirectory of test-center.  You can store the expected sha256 hashes there too.

compiler/utils.ss
```diff
@@ -334,28 +334,87 @@
...
+        (load-shared-object libcrypto)
+        (let ([make-context (foreign-procedure "EVP_MD_CTX_new" () void*)]
+              [free-context (foreign-procedure "EVP_MD_CTX_free" (void*) void)]
+              [sha256-method ((foreign-procedure "EVP_sha256" () void*))]
+              [digest-init (foreign-procedure "EVP_DigestInit_ex" (void* void* void*) boolean)]
+              [digest-update (foreign-procedure "EVP_DigestUpdate" (void* u8* size_t) boolean)]
+              [digest-final (foreign-procedure "EVP_DigestFinal_ex" (void* u8* u32*) boolean)])
+          (lambda (pathname)
+            (let ([context (make-context)])
+              (when (zero? context)
```
dybvig: `(eqv? context 0)` is more appropriate here.  zero? is unnecessarily general, i.e., it is true for 0, +0.0, and -0.0.

compiler/utils.ss
```diff
@@ -334,28 +334,87 @@
...
+                (external-errorf "failed to allocate native SHA-256 context"))
+              (dynamic-wind
+                void
+                (lambda ()
+                  (unless (digest-init context sha256-method 0)
+                    (external-errorf "failed to initialize native SHA-256 context"))
+                  (call-with-port (open-file-input-port pathname)
+                    (lambda (ip)
+                      (let ([buffer (make-bytevector 65536)])
```
dybvig: I think it's probably safe to read the entire file with `get-bytevector-all`:
```
(call-with-port (open-file-input-port pathname) get-bytevector-all)
```
and make only one call to `digest-update`.

compiler/utils.ss
```diff
@@ -334,28 +334,87 @@
...
+                        (let ([hash (substring stdout 0 64)])
+                          (if (andmap hex-digit? (string->list hash))
+                              (string-downcase hash)
+                              (try command* (cons (format "~a produced unexpected output: ~a" command stdout) rfailure*))))
+                        (try command* (cons (format "~a produced unexpected output: ~a" command stdout) rfailure*)))
+                    (try command* (cons (format "~a failed with message ~a" command stderr) rfailure*))))))))
+
+    (define sha256-file-implementation
+      (delay
+        (let ([libcrypto (getenv "COMPACT_LIBCRYPTO")])
```
dybvig, line 445: I would like to see this try to find a libcrypto if `COMPACT_LIBCRYPTO` isn't defined so users don't have to set the environment variable to get a native implementation.  On OSX, it should use the built-in Common Crypto library so people don't have to install `openssl`.  It looks like `CC_SHA256` is the way to go there.  You can just look to see if `/usr/lib/system/libcommonCrypto.dylib` exists rather than trying some other way to figure out if the process is being run under OSX.  For Linux, it should look in the commonly used locations for libcrypto.so.  We might eventually want the `Copmpact` tool to try to install libcerypto on Linux systems if it is not already installed.

compiler/utils.ss
```diff
@@ -334,28 +334,87 @@
...
+              [free-context (foreign-procedure "EVP_MD_CTX_free" (void*) void)]
+              [sha256-method ((foreign-procedure "EVP_sha256" () void*))]
+              [digest-init (foreign-procedure "EVP_DigestInit_ex" (void* void* void*) boolean)]
+              [digest-update (foreign-procedure "EVP_DigestUpdate" (void* u8* size_t) boolean)]
+              [digest-final (foreign-procedure "EVP_DigestFinal_ex" (void* u8* u32*) boolean)])
+          (lambda (pathname)
+            (let ([context (make-context)])
+              (when (zero? context)
+                (external-errorf "failed to allocate native SHA-256 context"))
+              (dynamic-wind
```
dybvig: It would be better to use `with-exception-handler` instead of `dynamic-wind` to free the context.  For a model, see the `passes.ss` code that deletes all of the files created by the compiler when an error occurs.  `dynamic-wind` is more appropriate when doing multitasking with continuations or engines, while `with-exception-handler` is specifically about handling error situations.  The only downside of the `with-exception-handler` solution is that you have to free the context on success as well, because the exception handler is called only when an exception is raised.

flake.nix
```diff
@@ -78,6 +78,9 @@
           nodejs = final.nodejs_latest;
         });
         isDarwin = pkgs.lib.hasSuffix "-darwin" system;
+        libcrypto =
```
dybvig: If we use the Common Crypto library on OSX (Darwin) we don't have to bring in libcrypto or set `COMPACT_LIBCRYPTO`.

dybvig (changes requested):
I've requested a few changes, but overall this looks pretty good to me.

compiler/test.ss
mhounslow: Good call — I addressed that in df9b3bd, so now the sha-256 inputs live under `test-center/sha256-test`, and each expected digest stored in a .sha256 file.

compiler/utils.ss
mhounslow: Still familiarizing myself 😅  updated in df9b3bd

compiler/utils.ss
mhounslow: Good call, I also make sure to normalize the EOF result to an empty `bytevector` so empty-file hashing still works

df9b3bd

compiler/utils.ss
mhounslow: Done here df9b3bd :)

compiler/utils.ss
mhounslow: The flow now is that when `COMPACT_LIBCRYPTO` is unset, the implementation tries Common Crypto through `CC_SHA256`, then common Linux `libcrypto` names, and finally the existing external commands.

I decided to do a load probe instead of `file-exists?` for the Common Crypto because macOS can provide the library through the dyld shared cache even when that path is not visible to filesystem checks, but when COMPACT_LIBCRYPTO is explicitly set, we just use that exact library and report initialization failures. 

https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes#Known-Issues

> New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

flake.nix
mhounslow: df9b3bd

compiler/utils.ss
```diff
@@ -334,28 +334,130 @@
...
-                      (let ([hash (substring stdout 0 64)])
-                        (if (andmap hex-digit? (string->list hash))
-                            (string-downcase hash)
-                            (try command* (cons (format "~a produced unexpected output: ~a" command stdout) rfailure*))))
-                      (try command* (cons (format "~a produced unexpected output: ~a" command stdout) rfailure*)))
-                  (try command* (cons (format "~a failed with message ~a" command stderr) rfailure*))))))))
+  ;; verifier key byte-for-byte. Hash in-process through Common Crypto or OpenSSL when available,
+  ;; retaining the external-command implementation as a fallback.
+  (module (sha256-file)
+    (define (bytevector->hex bytes)
```
dybvig, line 340: Just FYI, we could replace all of the generic arithmetic operators here with fixnum operators. The general rule is if a structure fits in memory, its size is a fixnum.  So you can always use fixnum operators for bytevector indices.

`+` => `fx+`, `=` => `fx=`, `*` => `fx*`, `bitwise-arithmetic-shift-right`=> `fxsrl`, `bitwise-and` => `fxand`.

This will have only minimal impact on speed unless we someday compile the Compact compiler at optimize-level 3.  Even then this is probably not a significant part of the overhead compared to reading the file and computing the hash.

In fact, we could probably get away with replacing all of this with just:
```
(define (bytevector->hex bytes) (format "~(~{~2,'0x~}~)" (bytevector->u8-list bytes)))
```

compiler/utils.ss
```diff
@@ -334,28 +334,130 @@
...
+            ((= i (bytevector-length bytes)) hash)
+          (let ([byte (bytevector-u8-ref bytes i)])
+            (string-set! hash (* i 2)
+              (string-ref hex-digits (bitwise-arithmetic-shift-right byte 4)))
+            (string-set! hash (+ (* i 2) 1)
+              (string-ref hex-digits (bitwise-and byte #xf)))))))
+
+    (define (read-bytevector pathname)
+      (let ([bytes (call-with-port (open-file-input-port pathname) get-bytevector-all)])
+        (if (eof-object? bytes) (make-bytevector 0) bytes)))
```
dybvig: `(make-bytevector 0)` => `#vu8()`

compiler/utils.ss
```diff
@@ -334,28 +334,130 @@
...
+              (let ([hash
+                     (with-exception-handler
+                       (lambda (c)
+                         (free-context context)
+                         (raise-continuable c))
+                       (lambda ()
+                         (unless (digest-init context sha256-method 0)
+                           (external-errorf "failed to initialize native SHA-256 context"))
+                         (let ([bytes (read-bytevector pathname)])
+                           (unless (digest-update context bytes (bytevector-length bytes))
```
dybvig: I believe you can just the call to `digest-update` when `get-bytevector-all` returns `#!eof`.

compiler/utils.ss
```diff
@@ -334,28 +334,130 @@
...
+                             (external-errorf "failed to update native SHA-256 digest")))
+                         (let ([digest (make-bytevector 32)])
+                           (unless (digest-final context digest #f)
+                             (external-errorf "failed to finalize native SHA-256 digest"))
+                           (bytevector->hex digest))))])
```
dybvig: If you move `free-context` call from below to just above the `bytevector->hex` call, you don't need the `(let ([hash ---]) ---)` wrapper.

flake.nix
```diff
@@ -224,6 +225,7 @@
             ];
 
             CHEZSCHEMELIBDIRS = "compiler::obj/compiler:third_party/compiler::obj/third_party/compiler:${nanopass}::obj/nanopass:${rough-draft}/src::obj/rough-draft:srcMaps::obj/srcMaps::obj/compiler";
+            COMPACT_LIBCRYPTO = libcrypto;
```
dybvig, line 228: Won't this effectively set `COMPACT_LIBCRYPTO` to the empty string on Darwin rather than leaving it unset?  If so, `(getenv "COMPACT_LIBCRYPTO")` will return an empty string rather than `#f`.

compiler/utils.ss
```diff
@@ -334,28 +334,130 @@
...
+                          (if (andmap hex-digit? (string->list hash))
+                              (string-downcase hash)
+                              (try command* (cons (format "~a produced unexpected output: ~a" command stdout) rfailure*))))
+                        (try command* (cons (format "~a produced unexpected output: ~a" command stdout) rfailure*)))
+                    (try command* (cons (format "~a failed with message ~a" command stderr) rfailure*))))))))
+
+    (define sha256-file-implementation
+      (delay
+        (let ([libcrypto (getenv "COMPACT_LIBCRYPTO")])
+          (if libcrypto
```
dybvig: It looks to me like `(getenv "COMPACT_LIBCRYPTO")` will return the empty string rather than false on OSX (see comment in flake.nix below).  If so, this needs to be `(if (and libcrypto (not (equal? libcrypto ""))) ---)`

compiler/utils.ss
dybvig: Interesting.  I'm glad you caught that.

dybvig (changes requested):
Mostly there, just a few more questions and suggestions.

compiler/utils.ss
mhounslow: That's much better :) 

b08e024

compiler/utils.ss
mhounslow: Now converts `#!eof` to `#vu8()`

b08e024

compiler/utils.ss
mhounslow: Yeh good point, `get-bytevector-all` returns `#!eof`, the opensll path will skip `digest-update` and finalizes the empty digest directly

b08e024

compiler/utils.ss
mhounslow: Thanks :) `free-context` now runs before `bytevector->hex`

b08e024

compiler/utils.ss
mhounslow: Yeh good catch...I added that empty-string check in b08e024

flake.nix
mhounslow: This was a useful exercise. I tested this on `aarch64-darwin`

```sh
nix eval --accept-flake-config --json \
  .#devShells.aarch64-darwin.default.COMPACT_LIBCRYPTO
```

^this evaluates to `null`, I then also in the dev shell directly

```sh
nix develop --accept-flake-config \
  --command /usr/bin/printenv COMPACT_LIBCRYPTO
```

which returned no value, which suggests to me that Nix is omitting `COMPACT_LIBCRYPTO` rather than exporting an empty string.

dybvig: If `printenv` is set to the empty string, would we expect the same behavior?   Is there a way to check the exit status of the call to /usr/bin/printenv?  It should be 0 if `COMPACT_LIBCRYPTO` is defined and 1 if it is not.

dybvig: In any case, the empty-string check you to the RHS of the definition of sha256-file-implementation should mean the right thing happens either way.

dybvig (changes requested):
Please just add a few additional test cases: one with multiple lines and "lf" line endings, one with multiple lines and "crlf" line endings (you might have to specify that the "crlf" line endings should be retained for this one in the .gitattributes file), and one that contains random binary data.  I don't know if we documented the fact, but we been using binary rather than text mode.  In particular, the sha256 hash for the mulitiline file with "crlf" endings should differ from the hash for an otherwise identical one with "lf" endings.

flake.nix
mhounslow: Yeh good point the status, so my initial output check was not sufficient by itself, for example when I ran this

```sh
nix develop --accept-flake-config \
  --command /usr/bin/printenv COMPACT_LIBCRYPTO

printf 'exit_status=%s\n' "$?"
```

It returned `exit_status=1`, with `COMPACT_LIBCRYPTO` being unset. Explicitly setting it to an empty string prints a blank line and returns `0`.

dybvig (approved):

PR #700, opened 2026-08-07
coips: add draft on a language-agnostic contract representation

RomarQ:
The CoIP proposes a second compiler output behind a flag: `compiler/analyzed-ir.sexp`. It holds the analyzed program (`Lloweredemit`) in the compiler's own vocabulary. `langs.ss` is the grammar, so the CoIP defines no schema of its own. Each ledger operation and each `emit` carries its expanded Impact VM instructions, which a plain unparse drops. An SDK in any language can then execute circuits with a small interpreter, instead of embedding a JavaScript engine.

Rendered document: https://github.com/RomarQ/compact/blob/coip/standard-contract-representation/coips/coip-xxxx.md

The draft follows MPS-0022, which states the problem and leaves the format open: midnightntwrk/midnight-improvement-proposals#188

## Implementations

[LFDT-Minokawa/compact#722](https://github.com/LFDT-Minokawa/compact/pull/722) implements the in-tree `--analyzed-ir` output described by this CoIP. The compiler owns the extraction pass and writes `compiler/analyzed-ir.sexp` when the flag is present.

[LFDT-Minokawa/compact#726](https://github.com/LFDT-Minokawa/compact/pull/726) implements the alternative `--ir-hook` approach. A Scheme hook receives the compiler stages and controls its own output. This provides a general extension point but does not define one artifact format.


RomarQ:
@kmillikin @pataei @dybvig @JosephDenman

Following our discussion in the [August 12 community call](https://github.com/LFDT-Minokawa/compact/pull/707#issuecomment-5278435213), I updated the proposal and prepared two implementations.

## Implementations

[LFDT-Minokawa/compact#722](https://github.com/LFDT-Minokawa/compact/pull/722) adds the `--analyzed-ir` output described by this CoIP. The compiler extracts the analyzed IR and writes it to `compiler/analyzed-ir.sexp`.

[LFDT-Minokawa/compact#726](https://github.com/LFDT-Minokawa/compact/pull/726) adds the alternative `--ir-hook` approach. A Scheme hook receives the compiler stages and controls its output. This approach provides a general extension point but does not define a standard artifact format.

Which approach do you think is more suitable for the compiler?

PR #707, opened 2026-08-11
August 12 community call agenda

pataei:
Add what you'd like to discuss to `meetings/2026-08-12.md` agenda. If you're not able to edit directly please leave a comment on this PR.


DavidM-D:
Would it be possible to add a link to #709 in `- [ ] **Binary compatibility** what guarantee can we make about`

kmillikin (approved):

dybvig (approved):

RomarQ:
# Topic: Language-Agnostic Contract Representation (https://github.com/LFDT-Minokawa/compact/pull/700)

**Participants:** Parisa Ataei, Kent Dybvig, Kevin Millikin, Joseph Denman, Rodrigo Quelhas, David, Bernard, Matthew, Ricardo, Simon Gellis, Edward Alvarado

## Decisions

- **Standardization timing**: No official, versioned intermediate representation for now. The project is at too early a level of maturity; requirements come first from people who use the existing IR, then the team can improve it and consider standardizing it as officially supported.
- **Artifact placement**: The representation is a separate compiler output behind a flag, not an extension of `contract-info`. Parisa wants separation of concerns, because contract-info is mainly for internal compiler use and she expects problems later if it carries this much information for other use cases.

## Action Items

- [ ] **Rodrigo Quelhas** — Pick which of the options discussed fits the Rust SDK need, then open a pull request with the proposal and implementation so the Compact team can review it and steer the long-term direction
- [ ] **TBD** — Surface the existing stop-after-pass hook through the `compactc` command line, and decide how to output the intermediate language code
- [ ] **Kevin Millikin** — Document how a contributor becomes a maintainer, based on what other LFDT projects at a similar life-cycle stage do (Kevin names maintainer status as the route to an official, supported backend)

## Discussion Points

### Which IR the representation is based on

- Kent framed two ends of the range: a representation of a compiler intermediate language taken before the split into TypeScript and circuit code, or ZKIR grown into something complete.
- ZKIR today cannot represent a whole contract. It omits constructors, and it leaves out the ledger operations that TypeScript can use but contracts cannot.
- Kent argued ZKIR is too low level to feed arbitrary backends. By that point loops are unrolled and everything is inlined, so a small contract becomes a huge ZKIR program that is expensive to process, and generated JavaScript would not resemble the current JavaScript output. He expects the same problem when ZKIR is used for cross-contract calls.
- Joseph favored extending ZKIR into that executable representation, because the team already considers doing it anyway. His position is to have as few executable formats as possible: matching JavaScript semantics to ZKIR semantics is already hard, and a third executable representation adds another consistency problem.
- Kevin wrote in chat that he no longer thinks this should be ZKIR. Joseph wrote that the Scheme compiler could target a final non-Scheme IR, with ZKIR and its optimizations as just another backend.
- Kevin suggested using the no-disclose IR as it stands. It is fairly stable but unversioned, so consumers must absorb changes such as new foreign curves and foreign fields when they move to a new Compact commit.
- Kent offered informal versioning: record the shape of the intermediate language at your last build and compare it, so you can always tell if it changed.

### What the Rust SDK needs

- Rodrigo's ask is not a new backend. He wants the compiled artifact extended so each circuit includes its body, so an SDK in any language can interpret it if it has an interpreter.
- A Rust SDK at Shielded already uses the no-disclose IR with an interpreter that builds the transcript before the ZK part, on Kevin's earlier suggestion, because ZKIR does not carry everything the SDK needs.
- The goal is to drop the current fork and get the solution upstream, so a Python SDK, for example, could write its own interpreter and call contracts.
- Rodrigo shared what his proof of concept produces today: https://github.com/Moonsong-Labs/midnight-rs/blob/main/tests/fixtures/compiled/election/compiler/contract-info.json
- Kent noted the constructor must also be included for full interpretation.
- Ricardo wrote in chat that contract-info was originally created as the equivalent of Solidity's ABI, so he sees no strong reason the representation could not fit there.

### How the compiler would expose it

- Kent proposed a hook that captures the output of a chosen pass, so a consumer asks the compiler only for the intermediate language form in no-disclose and then writes their own processor. This commits the project to no versioning, and the consumer can put a version in the pass they add.
- Kevin proposed a flag that stops the compiler after a chosen pass, which he finds useful for testing anyway, plus the ability to dump a binary representation of IR forms instead of S-expressions, with a specification so consumers can deserialize it.
- Kent said the plumbing already exists, because unit tests run only through a named pass such as infer types. The work is surfacing that hook through the `compactc` command line and deciding how to output the intermediate language code, which he called fairly straightforward.
- Kevin said JSON is the wrong format, and a binary representation that is fast to deserialize is the right way for a backend written outside Scheme.
- Kent flagged the hard part: pulling out all the information embedded in the intermediate language, such as VM code that sits in the middle of other structures.
- On user experience, Kevin said an SDK can wrap the compiler any way it wants. A shell script that invokes the compiler is fine, and users should not face a five-step process. He described it as the Unix philosophy, where a simple tool produces an output that feeds another tool.
- This mechanism will not serve cross-contract calls. Kevin said that design is already under way with many requirements and will not necessarily use this mechanism, but if cross-contract calls land in full in half a year, Rodrigo could re-architect on top of them.

### Ownership and the path to an official backend

- Kevin read the proposal as wanting an unofficial Rust backend outside the Minokawa project, written outside Scheme, which is why a JSON representation was requested.
- Kevin's alternative is to contribute it inside the project. The contributor becomes a maintainer, the TSC agrees to support the backend, and the contributor can then request or add things in the compiler front end.
- Kevin noted the cost: if the project takes it, the project commits to maintaining it, or eventually to deprecating and removing it.
- Parisa's takeaway is that the project does not want to own another backend to maintain at this time, which makes maintainer status the route for whoever builds it.
- Kevin wrote in chat that people will want to use this as a distribution format if it becomes an official one.

## Open Questions

- Which intermediate representation the artifact is built on. Rodrigo has several options and must pick one; the group did not choose for him.
- Whether NanoPass IR can be serialized, read back in a separate process, and then continued. Kevin raised it and suggested upstreaming it to NanoPass. Kent said FASL output works only if no Scheme procedures are embedded in the IR, and he does not remember whether natives and ledger forms still carry them, so it may need re-architecture.
- The output format. Kevin holds that JSON is wrong and binary is right; the working proof of concept emits JSON, and Ricardo argued in chat for the contract-info/ABI path.

PR #708, opened 2026-08-11
Fix for a pair of bugs in ZKIR v3 types nested in Compact types

kmillikin:
When a ZKIR v3 native type was nested inside a Compact type, the ZKIR v3 backend did not correctly handle the ledger encoding when emitting Impact instructions. This could lead to a `cannot-happen` assertion failure in the implementation.

Testing the fix for that issue revealed that the secp256k1 curve and field types could not be properly nested inside Compact types used in the ledger, because decoding the ledger representation did not properly consume the encoded input for subsequent decoding operations.


compiler/zkir-v3-passes/reduce-to-zkir.ss
```diff
@@ -311,12 +313,17 @@
...
+                       (circuit-alignment-for default-src
+                         (zkir-val-alignment* rand)
+                         (zkir-val-input* rand)
+                         (zkir-instr*))])
+           (zkir-instr* instr*)
+           (let ([code* (fold-left (lambda (code* a)
+                                     (cons (assemble-alignment-atom a) code*))
+                          (cons (length alignment*) code*)
+                          alignment*)])
+             (fold-left (lambda (code* var) (cons var code*)) code* var*)))]
```
dybvig: Is this just `(append (reverse var*) code*)`?

I would be tempted to write it that way.  The compiler **should** produce the same or better code (even though I don't think it does), and that's the criteria I use in all but super performance-critical code.

compiler/zkir-v3-passes/reduce-to-zkir.ss
```diff
@@ -562,85 +521,24 @@
...
-                               (cons
-                                 (make-public-input test-val var-name "Scalar<BLS12-381>")
-                                 (zkir-instr*))))
-                           var-name*)
-                         (values (map assemble-alignment-atom alignment*) var-name*))))])
+           ;; The public inputs are not yet bound in ZKIR.  Insert `public_input` instructions for
+           ;; them.
+           (assertf (= (length var-name*) (length primitive-type*))
+             "mismatched lists of public input names and types")
+           (let loop ([var-name* var-name*] [primitive-type* primitive-type*])
```
dybvig: This looks like a `for-each` written out:
```
(for-each
  (lambda (var-name primitive-type)
    (zkir-instr*
      (cons
        (make-public-input test-val var-name (type->string primitive-type))
        (zkir-instr*))))
  var-name*
  primitive-type*)
```

dybvig (approved):
The comment about code duplication on line 38 of `reduce-to-zkir.ss` is out-of-date and can probably just be removed (yay!).

I've suggested a couple of code simplifications as well.

compiler/test.ss
```diff
@@ -92359,6 +92359,69 @@ groups than for single tests.
...
+      "  const ps2 = pointStructs.lookup(2);"
+      "  jubjubMap.insert(0, default<Map<Uint<8>, JubjubPoint>>);"
+      "  jubjubMap.lookup(0).insert(1, ecMulGenerator(j0));"
+      "  const jp1 = jubjubMap.lookup(0).lookup(1);"
+      "}"
+      )
+    (stage-javascript
+      '("test('ZKIR native types in ledger ADTs', async () => {"
+        "  const [contract, context] = await startContract(contractCode, {}, 0);"
+        "  expect((await contract.circuits.test(context, 3n, 4n)).result).toEqual([]);"
```
JosephDenman: This asserts result equals `[]`, and the reads are bound (`const st`, `const ps`, `const jt`) but never compared — so a `fromValue` that had stopped throwing while still not consuming would return the first element three times and pass this test. Could you compare them against what was written? I think element-wise comparison is what actually pins this bug, since the failure mode we'd like to guard against is silently wrong data rather than an exception.

JosephDenman (changes requested):
LGTM. Just one request on the new test case.

compiler/zkir-v3-passes/reduce-to-zkir.ss
kmillikin: Yeah.

This whole thing is just:
```scheme
(append (reverse var*)
        (reverse (map assemble-alignment-atom alignment*)
        (cons (length alignment*) code*))
```

And note that `circuit-alignment-for` computes the lists in the order we want and then reverses them.  It leaks computation.  Maybe I'll clean that up in a separate PR.

compiler/zkir-v3-passes/reduce-to-zkir.ss
kmillikin: Yeah, I'll change it.  I do like named let, though.

compiler/test.ss
kmillikin: I'll add that.

kmillikin: I added JS tests that the ledger values that were written are the ones that we expect and also asserts in the contract that the written and read values are Compact `==` to connect the in-circuit values with the public inputs in the transcript.

JosephDenman (approved):
LGTM

PR #711, opened 2026-08-11
compiler: fix incorrect comments in bytes->field guard derivation

Jc-asastu:
In `missing-guard-workarounds.ss`, the comments describing the `bytes->field` guard say `triv1 == 0`, but the temporary they refer to is defined a few lines above as `triv1 == q`:

```scheme
;; t2 = triv1 == q
`(= 1 ,t2 (== ,triv1 ,q))
...
;; t4 = !(triv2 > r) && triv1 == 0    <- t4 uses t2, i.e. triv1 == q
`(= 1 ,t4 (select ,t3 0 ,t2))
```

The generated code is correct; only the comments are off. This is comment-only — no behaviour change.

Worth fixing because this pass is the workaround for zkir's non-conditional operators, and the inline derivation is the main aid for anyone auditing that arithmetic.

### DIFF
--- a/compiler/circuit-passes/missing-guard-workarounds.ss
+++ b/compiler/circuit-passes/missing-guard-workarounds.ss
@@ -124,12 +124,12 @@
                         `(= 1 ,t2 (== ,triv1 ,q))
                         ;; t3 = triv2 > r
                         `(= 1 ,t3 (< ,(unsigned-bits) ,r ,triv2))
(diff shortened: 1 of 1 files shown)


Jc-asastu:
Note: this contribution was AI-assisted. Every change was verified against the source before submitting. Happy to have the `ai-assisted` label applied if that's your convention — I can't set labels as an outside contributor.

PR #714, opened 2026-08-12
Dynamic cross-contract calls - Q3

JosephDenman:
# Dynamic cross-contract module resolution

This is CoIP-3. Callers no longer import the callee's module at compile time - the application supplies a provider, and the runtime resolves the module by address at call time and checks it before entering it. The practical upshot is that deploying a new implementation at an address doesn't require recompiling everything that calls it.

For a detailed description of the changes, see [‎CHANGELOG.md‎](https://github.com/LFDT-Minokawa/compact/pull/714/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed). 

Note [CoIP-3](https://github.com/jonathan-sobel/compact/blob/jsobel/coip-ccc-dynamic-code/coips/coip-xxxx.md) is expected to capture this work. It's still a work in progress, so there might be temporary divergences.


runtime/src/contract.ts
```diff
@@ -14,101 +14,202 @@
...
+    failResolution(resolutionContext, { kind: 'UnsupportedImplementation' });
+  }
+  if (typeof thunk !== 'function') {
+    failResolution(resolutionContext, { kind: 'ProviderThrew', cause: thunk });
+  }
+  let calleeModule: Module;
+  try {
+    calleeModule = await thunk();
+  } catch (cause) {
+    failResolution(resolutionContext, { kind: 'ModuleLoadRejected', cause });
```
pataei, line 201: I'm not sure if the scenario would happen where the following condition is met but if it does it won't throw a `ModuleResolutionFailure` (I think it'd throw a bare `TypeError` naming neither the contract nor the call) so adding the following check before the `missing` filter might prevent that case:

```
if (typeof calleeModule !== 'object' || calleeModule === null) {
    failResolution(resolutionContext, { kind: 'ProviderThrew', cause: calleeModule });
  }
```

runtime/src/providers.ts
```diff
@@ -14,15 +14,28 @@
...
+/** A deferred load of a generated contract module: evaluated only when a call resolves to it. */
+export type ModuleThunk = () => Promise<Module>;
+
+/**
+ * A user-provided lookup from a cross-contract callee's address to the module implementing the
+ * contract deployed there.
+ *
+ * `resolve` is synchronous and total: loading is deferred into the thunk, and an address with no
+ * binding returns `undefined` rather than throwing, so the runtime classifies every failure and an
+ * application sees one vocabulary rather than one per provider.
```
pataei: Change 
"so the runtime classifies every failure and an application sees one vocabulary rather than one per provider." 
to 
"so the runtime classifies every failure that isn't caught at compile time and an application sees one vocabulary rather than one per provider.*

Reason: if a declared circuit isn't provided the compiler complains in a non-`ModuleResolutionFailure` error.

compiler/typescript-passes/print-typescript.ss
```diff
@@ -1964,169 +1935,132 @@
...
+            xpelt*)))
+
+      ;; contract-type* holds exactly the contract types a call is made on.
+      (define (print-declared-interfaces contract-type*)
+        (print-table "declaredInterfaces"
+          (map
+            (lambda (contract-type)
+              (nanopass-case (Ltypescript Contract-Type) contract-type
+                [(tcontract ,src ,contract-name (,elt-name* ,pure-dcl* (,type** ...) ,type*) ...)
+                 (format "~a: ~a"
```
pataei, line 2071: What happens if two imported modules both declare `contract C`? If `C` as a string is added to the table there will be a conflict and I cannot remember of the top of my head is this is caught/warned in `expand-modules-and-types` or if the `contract-name` is prefixed with the module name.

runtime/src/contract.ts
```diff
@@ -462,39 +501,87 @@ const forbiddenCalleeWitnesses = (calleeAddress: ocrt.ContractAddress): Record<s
...
+    assertDefined(deployedState, `deployed contract state for callee '${calleeAddress}'`);
+
+    // 6. The module, from the provider.
+    const calleeModule = await resolveModule(moduleProvider, calleeAddress, resolutionContext);
+
+    // 7. Conformance before keys: a non-conformant module is an application mistake, a key mismatch
+    //    is the code and the chain having drifted.
+    checkModuleConformance(declaration, calleeModule, resolutionContext);
+
+    // 5 and 8. The operation exists and carries a key, and its fingerprint agrees.
```
pataei: I'm confused about the order of operations here based on the comments.

runtime/src/contract.ts
JosephDenman: Yep. Right on. Fixed.

runtime/src/providers.ts
JosephDenman: Ok. Updated the comment. Hopefully more clear now.

compiler/typescript-passes/print-typescript.ss
JosephDenman: This is caught in `expand-modules-and-types` via the `env-insert!` function. It refuses a second, different binding for the same symbol in the same scope. So, `print-declared-interfaces` never sees a duplicate key.

dybvig (commented):
I looked over the compiler changes and didn't spot anything.  Since I'm not intimately familiar with the runtime, I gave Claude a rather vague request: Can you tell me if the changes seem to do what they are intended to do and if there are any missing cases?  Here's what it came up with.  I looked a bit into 1 and 2, and both look like real issues (the offending code in 2 existing already in `main`).  I did not look closely into 3 and 4.  5 is probably just a comment fix.

1. "A declared-impure circuit may be implemented pure" is asserted but not implemented. conformance.ts:292 reads if (!declared.pure && !implemented.provable) → Provability. A pure circuit is never provable — the compiler's provable flag is membership in the impure export list — so a declared-impure circuit implemented pure always fails. The test at conformance.test.ts:323 that claims otherwise passes only because it hand-builds sig(true, true, …), a pure && provable signature compactc cannot emit; their own compiler test ('pure and provable are independent') exercises exactly the three reachable combinations and not that one. Either the rule wants to be !declared.pure && !implemented.provable && !implemented.pure, or the test and its comment should go. As written the file documents a policy the code doesn't have.

2. pure in a contract T { } block is accept-and-fail. desugar-contract-calls.ss carries pure-dcl* through synth-tc-native! and extend-tcontract without ever consulting it, so a.peek() on a declared-pure circuit desugars into the full transientCommit/claimContractCall machinery and then dies at crossContractCall step 1 with PureInterfaceCircuit. Nothing rejects it at compile time. checkConformance will happily validate such a circuit first, which makes it look supported. This isn't a regression — main reached assertDefined(provableCircuit, …) and failed more obscurely — but the branch adds a named failure for it rather than a diagnostic, and Module.pureCircuits is now declared and read nowhere in runtime/src (and is absent from RESOLVED_MODULE_EXPORTS, so it isn't even required to exist). Worth deciding: compile error, or dispatch through pureCircuits.

3. The new failure paths mutate the context before validating it. crossContractCall runs resolveQueryContext (step 4) before resolveModule, checkModuleConformance and checkImplementation (steps 6–8), so all three throw after queryContexts, gasCosts and contractStates have been populated. Main had the same shape, but the branch adds cached.effects = emptyEffects() on the repeat-call path (contract.ts:251) — so a second call to an already-resolved callee now discards that callee's accumulated effects and then may throw before anything re-populates them. Since ModuleResolutionFailure is explicitly built as a catchable payload ("so it survives an application re-throwing through its own error type"), applications are being invited to catch and continue on a context that is no longer coherent. Splitting the state fetch from the query-context construction, and doing the three checks in between, would make failures non-destructive. (Relatedly, the step comments run 4, 6, 7, "5 and 8", 9 — the intended order is no longer readable.)

4. Untrusted signatures are tag-checked but not shape-checked. firstUnreadableTag is careful to read tag as string rather than trust it, and resolveModule hardens against a provider returning junk — but past the tag, t.elements.map, t.types, t.circuits and t.length are trusted. A module with {tag:'Struct'} and no elements throws a bare TypeError out of firstUnreadableTag, and {tag:'Vector', length:-1} throws RangeError from new Array(t.length) in sequenceElements — both escaping the ModuleResolutionError vocabulary the rest of the file works to maintain. sequenceElements also materializes both arrays before comparing lengths, so {tag:'Vector', length: 2**30} allocates a billion slots against a two-element tuple. Given everything else in this file assumes the callee module wasn't compiled by us, this is the one place the threat model isn't applied.

5. Nominal identity is a bare string across compilation units. Struct, Enum and Alias compare by name only; the doc-comment's justification — "those names are shared across compilation units through the module declaring them" — is an assumption nothing carries or enforces. Two independently authored contracts each with new type Meters = Uint<64> are indistinguishable, as are two contract types with the same circuit shapes (the contract name is deliberately not compared). So conformance validates shape, never identity: the file header's Vault/Token example is caught only because their shapes differ. That's probably the right trade-off, but it belongs in the header as a stated limit rather than implied by a comment that reads as if identity were preserved.

JosephDenman:
@dybvig Most of those are valid to some degree, even if they're comment issues. The most consequential is point two. The real problem with that is soundness. An impure cross-contract call is chain-validated: the callee gets its own transcript in the intent, `kernelClaimContractCall` binds its I/O into the caller's via `commCommData`, and the ledger replays it. A pure circuit has none of that. A malicious provider returns whatever it likes and the proof still verifies. So, I would argue we shouldn't allow cross-contract pure circuit invocations.

JosephDenman:
@dybvig I believe I've addressed all of your review comments.

compiler/analysis-passes/expand-modules-and-types.ss
```diff
@@ -841,6 +841,10 @@
                    (handle-fun src 'native pelt exported? function-name type-param*)]
                   [(witness ,src ,exported? ,function-name (,type-param* ...) (,arg* ...) ,type)
                    (handle-fun src 'witness pelt exported? function-name type-param*)]
+                  ;; TODO: reject a true pure-dcl here. A pure cross-contract call has no transcript,
```
dybvig, line 844: I don't understand the restriction.  For circuits declared pure, can we not push the call through a different path where it makes a normal function call through the pure-circuits object?

I do get that failing to declare a circuit pure that actually is pure is problematic.

dybvig (approved):

PR #719, opened 2026-08-18
Make 'CompactType' implementations and testing more robust

JosephDenman:
PR https://github.com/LFDT-Minokawa/compact/pull/708 included a fix to the FAB -> TypeScript decoding logic for `CompactType`. While I was debugging that issue, I found some other improvements we should make to `CompactType`. This PR contains those improvements. See the change log for details.


compiler/typescript-passes/print-typescript.ss
```diff
@@ -1127,12 +1127,26 @@
...
+                           (lambda (field bound)
+                             (format "typeof(~a.~a) === 'bigint' && ~a.~a >= 0n && ~a.~a <= __compactRuntime.~a"
+                               var field var field var field bound))])
+                     (nanopass-case (Ltypescript Curve-Type) ctype
+                       [(curve-jubjub)
+                        (format "~a && ~a"
+                          (coordinate "x" "MAX_FIELD")
+                          (coordinate "y" "MAX_FIELD"))]
+                       [(curve-secp256k1)
+                        (format "~a && ~a && typeof(~a.identity) == 'boolean'"
```
dybvig: In both the new and the old code, the `typeof` check for `bigint` uses `===` while the `typeof` check for `boolean` uses `==`.  Is there a reason for that, or should they both use `===`?

dybvig (commented):
LGTM, just a question about using `==` to test for `boolean` while we use `===` to test for `bigint`.

compiler/typescript-passes/print-typescript.ss
JosephDenman: No reason. In this case, they're the same. `==` just means that JS will try to coerce both operands to the same type before testing equality. In this case `typeof` always returns a string, so there's nothing to coerce. Still, fixed for consistency.

dybvig (approved):

kmillikin:
Hmmm, I don't like using `SECP256K1_LOW_LIMB_BOUND` and I would certainly not export it.  The encoding in ZKIR could certainly change and that should be a non-breaking change in the Compact runtime.  And I definitely don't want us to export a bunch more of these as we add new curve types.

The functions that use it are not fully abstracted anyway:

```javascript
  alignment(): ocrt.Alignment {
    return [
      { tag: 'atom', value: { tag: 'bytes', length: 24 } },
      { tag: 'atom', value: { tag: 'bytes', length: 8 } },
    ];
  },

  fromValue(value: ocrt.Value): bigint {
    if (value.length < 2 || value[0] == undefined || value[1] == undefined) {
      throw new CompactError('expected Secp256k1Base');
    }
    const limbs = value.splice(0, 2);
    const low192 = ocrt.valueToBigInt([limbs[0]]);
    const high64 = ocrt.valueToBigInt([limbs[1]]);
    if (low192 >= SECP256K1_LOW_LIMB_BOUND) {
      throw new CompactError('expected Secp256k1Base');
    }
    let res = high64 << 192n | low192;
    // The ZKIR representation subtracts 1 from the value.
    res = (res == MAX_SECP256K1_BASE) ? 0n : res + 1n;
    if (res > MAX_SECP256K1_BASE) {
      throw new CompactError('expected Secp256k1Base');
    }
    return res;
  },

  toValue(value: bigint): ocrt.Value {
    if (value < 0n || value > MAX_SECP256K1_BASE) {
      throw new CompactError('expected Secp256k1Base');
    }
    // The ZKIR representation subtracts 1 from the value.
    value = (value == 0n) ? MAX_SECP256K1_BASE : value - 1n;
    return ocrt.bigIntToValue(value & ((1n << 192n) - 1n))
      .concat(ocrt.bigIntToValue(value >> 192n));
  },
```
Includes multiple occurrences of 24, 32-24, and 8*24.  I'd just go ahead and directly use `1n << 192n - 1n` as it was before.

kmillikin:
Looking at this implementation, I don't know why we give the alignment tags `(bytes 24) (bytes 8)`, because the implementation of toValue and fromValue assumes that those are native field values.  Why not `(field) (field)`?

PR #730, opened 2026-08-19
CoIP: add a Rust code generation backend

yshyn-iohk:
Adds a draft CoIP proposing an additional, **opt-in** Rust code generation backend for the Compact compiler (`--rust`, with `--skip-ts`), plus a supporting `compact-runtime` Rust crate, so applications outside JS/TS can drive Compact contracts without embedding a JavaScript engine.

Per CoIP-1 this PR **describes** the change; it contains no implementation. A complete reference implementation is linked from the Implementation section.

### Status
- Document status: **Draft**; filename is `coips/coip-xxxx.md` pending TSC numbering.
- **I have deliberately not assigned the TSC chair as a reviewer yet** — per CoIP-1 that signal means "ready for numbering", and I would like to self-review and take any early feedback first. I will assign and comment when ready.

### Why this, and why here
This follows guidance given on the MPS-0022 review thread (midnightntwrk/midnight-improvement-proposals#188): *"we're certainly open to adding other supported target language backends. You would have to implement only a very small number of passes (the TypeScript backend) to target another language, and you would need some equivalent of the Compact runtime"*. The proposal reuses the existing front end and IRs and adds only emission passes — no new IR, no new language semantics, and no change to the TypeScript path when `--rust` is absent.

### Points where TSC direction would be most useful
1. **Category.** Filed as `Tooling`; `Language` may be a better fit since it defines the API of generated code (as CoIP-1 describes for generated JavaScript). Happy to recategorize.
2. **Where the runtime crate lives.** Proposed in-repo, symmetric with `runtime/` (`@midnight-ntwrk/compact-runtime`) and `tests-e2e/`, so the compiler↔runtime version guarantee provided by `runtime/export-version.ss` extends to the Rust runtime. A separate repository is a viable alternative and is listed under Rejected Ideas — the authors are willing to do either; it changes packaging only.
3. **Relationship to MPS-0022 / #700.** The Rejected Ideas section positions ahead-of-time code generation as complementary to ZKIR/IR interpretation rather than an alternative to it. If the TSC sees these as competing, that is worth settling early.

### Reference implementation
Linked in the document. It is validated by a byte-parity harness against the TypeScript backend plus executing end-to-end tests, and has been exercised on real contracts (a 19-ledger-field DID contract resolved live against a standalone network, and most of the Midnight verifiable-credentials contract suite). Its currently released line targets ledger 8; a branch tracking this repository's current ledger-9 rc pin is in progress.


yshyn-iohk:
Updated the draft: added @patextreme as a co-author, and reworked the document for readability — a concrete compile-and-use example up front, tables replacing dense prose for the emitted layout and the runtime-placement comparison, shorter sections, and the correctness argument stated in plain terms. No change to the proposal itself.

Still deliberately **not** requesting TSC numbering yet (no reviewer assigned, per CoIP-1) — will signal when the authors are ready.

patextreme:
Two pieces of interface feedback, best settled while the document is still Draft:

**1. Prefer `--target <language>` over `--rust` + `--skip-ts`**

- CLI flags are API. A boolean `--rust` assumes exactly two backends and forces a breaking rename when a third language arrives — something the Motivation section itself anticipates.
- Suggested semantics: `--target` is **repeatable**; values `ts` and `rust`; the explicit list is authoritative, with `ts` implied only when the flag is absent:

  | Invocation | TS | Rust |
  |---|---|---|
  | `compactc counter.compact out/` | ✅ implied | — |
  | `compactc --target rust counter.compact out/` | — | ✅ |
  | `compactc --target rust --target ts counter.compact out/` | ✅ | ✅ |

- Since the list replaces the default, `--target rust` alone means Rust-only — subsuming `--rust --skip-ts`, with no flag-combination validation and "emit both" (needed by the parity harness) still expressible.
- Space-separated, matching existing value flags (`--compact-path`, `--sourceRoot`); unknown values error, listing the valid ones.
- Keep ZKIR/keys orthogonal (`--skip-zk` unchanged) and #700's `analyzed-ir` out of this flag — it's an artifact, not a contract-code backend.

**2. Name the crate `midnight-compact-runtime`, not `compact-runtime`**

- The Midnight Rust ecosystem is consistently `midnight-*` (`midnight-onchain-state`, `midnight-transient-crypto`, …). The TS runtime's npm scope serves this purpose; the Rust analogue is the `midnight-` prefix.
- The name is generic on crates.io and generated code pins it — effectively permanent after first release, trivial to fix now.

Both are naming/surface decisions; neither changes the design.

yshyn-iohk:
Thanks, @patextreme, for your comments.

Regarding [1], it looks better from a `perfection` point of view, but it will introduce backward compatibility issues when the new compiler version is released. 
The idea behind this improvement is to make the transition and integration as smooth as possible, so developers can use the new compact compiler without any changes to the semantics of the calls used to generate TS and ZKIR keys. 
That's why two flags: `--rust` + `--skip-ts` naturally extend the functionality of the compiler.
Also, this capability might be treated as `experimental`, till the full stabilization and wider adoption by the Minokawa community.

Regarding [2] - you are right; let me make these changes in our branches.

patextreme:
Thank you for considering both points.

On point 1, I appreciate the goal of a smooth transition, and `--target` preserves it: neither `--rust` nor `--skip-ts` exists today, and with `ts` implied when the flag is absent, existing invocations are untouched under either design.

The difference is forward compatibility: a second backend would make `--skip-ts` ambiguous and force the breaking rename that `--target` avoids. If the concern is scripts already using the reference implementation's flags, a deprecation alias can cover those.

Given that, I'd still encourage `--target` before the interface ships.

yshyn-iohk:
@patextreme, you are right. `--target` is more universal

Also, without `--target` the compact compiler will behave as it used to be

I created the issue in the fork and will deliver it soon.

PR #749, opened 2026-08-28
Implementation of the secp256r1 foreign curve

kmillikin:
This also includes a large amount of the infrastructure for the Curve25519 foreign curve, though it is not actually exported from the standard library yet.


compiler/analysis-passes/expand-modules-and-types.ss
```diff
@@ -301,14 +315,22 @@
                          elt-name1* pure-dcl1* type1** type1*))
                 elt-name2* pure-dcl2* type2** type2*))
       (define (same-curve-type? ctype1 ctype2)
-        (nanopass-case (Lexpanded Curve-Type) ctype1
+        (strict-nanopass-case (Lexpanded Curve-Type) ctype1
```
dybvig, line 318: Just something to remember: `strict-nanopass-case` should also work for `same-field-type?` next time we have to change the code.

compiler/analysis-passes/check-types-Lnodca.ss
```diff
@@ -268,6 +290,12 @@
          type]
         [else (assert cannot-happen)]))
      (define (arithmetic-binop src op result-type expr1 expr2)
+       (define (check-curve-type ctype)
```
dybvig: Rather than `check-curve-type`, which to me implies that an exception will be raised when the check fails, I'd rather use something like `is-zkir-v3-only-curve?` as in `infer-types`.

compiler/analysis-passes/infer-types.ss
```diff
@@ -989,13 +1014,13 @@
     [(native ,src ,function-name ,native-entry (,[arg*] ...) ,[Return-Type : type src "circuit" -> type])
      (build-function (native-entry-class native-entry) #t function-name arg* type)]
     [(witness ,src ,function-name (,[arg*] ...) ,[Return-Type : type src "witness" -> type])
-     (check-secp256k1 type)
+     (check-zkir-v3-curve type)
```
dybvig, line 1017: We should find a way not to have to place these checks wherever they are needed because it's fragile and a bit awkward.  Of course, they'll go away temporarily once we eliminate the v2 backend, but some similar checks might take their place when we add a v4 backend.

compiler/analysis-passes/infer-types.ss
```diff
@@ -1817,6 +1842,13 @@
            [(tunsigned ,src ,nat) (>= nat 255)]
            [(tfield ,src ,ftype) #t]
            [else #f]))
+       ;; For base and scalar fields, only some casts to/from Bytes are supported.
+       (define (check-length ctype len)
```
dybvig: Again, I would prefer a predicate name here rather than `check-*`.

compiler/analysis-passes/infer-types.ss
```diff
@@ -1817,6 +1842,13 @@
...
+         (strict-nanopass-case (Ltypes Curve-Type) ctype
+           [(curve-curve25519) (eqv? len 64)]
+           [(curve-jubjub) #f]
```
dybvig, line 1849: maybe this should be an assert cannot-happen?

compiler/circuit-passes/check-types-Linlined.ss
```diff
@@ -55,15 +55,29 @@
       (nanopass-case (Linlined Argument) arg
         [(,var-name ,type) type]))
     (define (format-field-type ftype)
-      (nanopass-case (Linlined Field-Type) ftype
+      (strict-nanopass-case (Linlined Field-Type) ftype
```
dybvig, line 58: If we aren't yet ready to commonize the types, we should at least consider using macros to abstract `format-field-type` and other operations over the language name.

compiler/circuit-passes/flatten-datatypes.ss
```diff
@@ -383,14 +405,23 @@
...
-           (assert (= (length triv*) 2))
+         (define (make-foreign-field-cast ctype)
            (hashtable-set! var-ht var-name (Wump-single var-name))
-           (list `(= ,test ,var-name (bytes->field ,src ,ftype ,len ,(car triv*) ,(cadr triv*)))))
-         (nanopass-case (Lflattened Field-Type) ftype
-           [(field-base (curve-secp256k1)) (make-secp256k1-cast)]
-           [(field-scalar (curve-secp256k1)) (make-secp256k1-cast)]
+           (strict-nanopass-case (Lflattened Curve-Type) ctype
+             [(curve-curve25519)
+              (assert (= (length triv*) 3))
```
dybvig, line 412: don't love these manifest constants (2 and 3 and earlier the 32 and 64): you're right that we need to abstract the handling of curves and fields where at least the 2 and 3 are derived from the 32 and 64 in relation to the field size.

compiler/circuit-passes/optimize-circuit.ss
```diff
@@ -148,11 +148,20 @@
       (define (mbits-hash mbits hc)
         (if mbits (bits-hash mbits hc) hc))
       (define (field-type-hash ftype hc)
-        (nat-hash (nanopass-case (Lflattened Field-Type) ftype
+        (nat-hash (strict-nanopass-case (Lflattened Field-Type) ftype
```
dybvig, line 151: our abstraction could even generate these hash seeds based on the relative order in which curves and fields are defined

compiler/typescript-passes/print-typescript.ss
```diff
@@ -890,16 +896,19 @@
...
+        [(curve-secp256k1) "Secp256k1"]
+        [(curve-secp256r1) "Secp256r1"]))
     (define (format-field-type ftype)
-      (nanopass-case (Ltypescript Field-Type) ftype
+      (strict-nanopass-case (Ltypescript Field-Type) ftype
         [(field-native) "Field"]
-        [(field-scalar (curve-jubjub)) "JubjubScalar"]
-        [(field-base (curve-secp256k1)) "Secp256k1Base"]
-        [(field-scalar (curve-secp256k1)) "Secp256k1Scalar"]))
+        [(field-base ,ctype) (format "~aBase" (format-curve-type ctype))]
```
dybvig, line 888: nice

compiler/typescript-passes/print-typescript.ss
```diff
@@ -1017,20 +1026,29 @@
...
-                 (nanopass-case (Ltypescript Field-Type) ftype
+                 (strict-nanopass-case (Ltypescript Field-Type) ftype
                    [(field-native) "__compactRuntime.CompactTypeField"]
-                   [(field-scalar (curve-jubjub)) "__compactRuntime.CompactTypeField"]
-                   [(field-base (curve-secp256k1))
-                    "__compactRuntime.CompactTypeSecp256k1Base"]
-                   [(field-scalar (curve-secp256k1))
-                    "__compactRuntime.CompactTypeSecp256k1Scalar"])]
+                   [(field-base ,ctype)
+                    (strict-nanopass-case (Ltypescript Curve-Type) ctype
```
dybvig: Can this be abstracted using `format-field-type`?

compiler/typescript-passes/print-typescript.ss
```diff
@@ -1017,20 +1026,29 @@
...
+                      [(curve-curve25519) "__compactRuntime.CompactTypeCurve25519Scalar"]
+                      [(curve-jubjub) "__compactRuntime.CompactTypeField"]
+                      [(curve-secp256k1) "__compactRuntime.CompactTypeSecp256k1Scalar"]
+                      [(curve-secp256r1) "__compactRuntime.CompactTypeSecp256r1Scalar"])])]
                 [(tunsigned ,src ,nat)
                  (format "new __compactRuntime.CompactTypeUnsignedInteger(~dn, ~d)"
                    nat (max 1 (byte-length nat)))]
                 [(tpoint ,src ,ctype)
-                 (nanopass-case (Ltypescript Curve-Type) ctype
+                 (strict-nanopass-case (Ltypescript Curve-Type) ctype
```
dybvig: Can this be abstracted using format-field-type?

compiler/typescript-passes/print-typescript.ss
```diff
@@ -1117,11 +1135,20 @@
                   [(tboolean ,src) (format "typeof(~a) === 'boolean'" var)]
                   [(tfield ,src ,ftype)
                    (let ([field-name
-                           (nanopass-case (Ltypescript Field-Type) ftype
+                           (strict-nanopass-case (Ltypescript Field-Type) ftype
                              [(field-native) "FIELD"]
-                             [(field-scalar (curve-jubjub)) "JUBJUB_SCALAR"]
-                             [(field-base (curve-secp256k1)) "SECP256K1_BASE"]
-                             [(field-scalar (curve-secp256k1)) "SECP256K1_SCALAR"])])
+                             [(field-base ,ctype)
```
dybvig, line 1094: Can this be abstracted using format-curve-type?
using something like `(format "~:@(~a~)_BASE" (format-curve-type ctype))`

compiler/typescript-passes/print-typescript.ss
```diff
@@ -2607,9 +2655,11 @@
              [(tfield ,src ,ftype) "0n"]
              [(tunsigned ,src ,nat) "0n"]
              [(tpoint ,src ,ctype)
-              (nanopass-case (Ltypescript Curve-Type) ctype
+              (strict-nanopass-case (Ltypescript Curve-Type) ctype
+                [(curve-curve25519) (assert cannot-happen)]  ;; TODO(kmillikin): implement.
```
dybvig: `cannot-happen` => `not-implemented`
(or add the implementation)

compiler/typescript-passes/print-typescript.ss
```diff
@@ -2731,10 +2781,22 @@
...
-                  ;; infer-types guarantees that these are the only possibilities.
-                  [(tfield ,src^ (field-native)) "addField"]
-                  [(tfield ,src^ (field-base (curve-secp256k1))) "secp256k1BaseAdd"]
-                  [(tfield ,src^ (field-scalar (curve-secp256k1))) "secp256k1ScalarAdd"]
+                  ;; `cannot-happen` below is guaranteed by `infer-types`
+                  [(tfield ,src^ ,ftype)
+                   (strict-nanopass-case (Ltypescript Field-Type) ftype
+                     [(field-native) "addField"]
+                     [(field-base ,ctype)
+                      (strict-nanopass-case (Ltypescript Curve-Type) ctype
```
dybvig, line 2699: might be more opportunities here and below to abstract using `format-field-type`

compiler/field.ss
```diff
@@ -30,13 +32,13 @@
...
-  ; relation to the proof system.  max-field is the largest representable field value.
-  ; WARNING: keep in sync with midnight-base-crypto. Will be caught by tests.
+  ;; a field value is a natural number whose range is bounded by a prime number determined in
+  ;; relation to the proof system.  max-field is the largest representable field value.
+  ;; WARNING: keep in sync with midnight-base-crypto. Will be caught by tests.
   (define (max-field) 52435875175126190479447740508185965837690552500527637822603658699938581184512)
   (define-field-predicate field? max-field)
 
-  ; This value should match `MAX_JUBJUB_SCALAR` in the Compact runtime.
+  ;; This value must match `MAX_JUBJUB_SCALAR` in the Compact runtime.
```
dybvig, line 41: Do we check this in the generated typescript?  I believe we do check the maximum field value.

compiler/zkir-v3-natives.ss
```diff
@@ -14,31 +14,64 @@
 ;;; limitations under the License.
 
 ;; ==== Non-native fields and curve points
+;; TODO(kmillikin): enable Curve25519 in the standard library when it is fully supported.
```
dybvig, line 17: I'm not sure there's any harm in exposing these even now.

compiler/zkir-v3-natives.ss
```diff
@@ -14,31 +14,64 @@
...
 (declare-native-entry circuit neg
   "__compactRuntime.secp256k1ScalarNeg"
   ([s (TypeRef Secp256k1Scalar) (discloses "the negation of")])
   (TypeRef Secp256k1Scalar))
 
-(declare-native-entry circuit inv
-  "__compactRuntime.secp256k1ScalarInv"
-  ([s (TypeRef Secp256k1Scalar) (discloses "the inverse of")])
-  (TypeRef Secp256k1Scalar))
+(declare-native-entry circuit neg
```
dybvig, line 43: The format of these leaves something to be desired, with the javascript runtime name preceding the type signature, but nothing we need to worry about now.

runtime/src/built-ins.ts
```diff
@@ -277,6 +302,41 @@ export function secp256k1ScalarAdd(x: bigint, y: bigint): bigint {
...
+
+// -- Negation
+/**
+ * Secp256k1 base field negation
+ *
+ * This function returns the negation of x in the secp256k1 base field.  That
+ * is, a value y such that x + y = 0 (modulo SECP256K1_BASE_MODULUS).  x is
+ * assumed to be in the range [0, SECP256K1_BASE_MODULUS).
+ */
+export function secp256k1BaseNeg(x: bigint): bigint {
```
dybvig, line 336: It seems like we could just use `secp256k1BaseSub(0n, x)` in the generated javascript code...but this is fine as is.

runtime/src/constants.ts
```diff
@@ -60,14 +60,25 @@ export const SECP256K1_SCALAR_MODULUS: bigint =
...
 /**
- * The exclusive upper bound of the low-order limb of a secp256k1 field value
- * in its field-aligned binary representation.
- *
- * A secp256k1 field value is 256 bits wide but a native field element holds
- * only 255, so it is split across two atoms: 192 low-order bits (24
- * bytes), then 64 high-order bits (8 bytes).
+ * The order of the secp256r1 base field
+ */
+export const SECP256R1_BASE_MODULUS: bigint = 2n**256n - 2n**224n + 2n**192n + 2n**96n - 1n;
```
dybvig, line 65: It would be nice to generate all or part of this file based on ../compiler/field.ss.

dybvig (approved):
I've marked a lot of stylistic things that might allow the code to be simplified or clarified as well as some notes for future consideration, but I didn't spot any problems.

compiler/analysis-passes/expand-modules-and-types.ss
kmillikin: I was assuming that there were no fields other than native, Base<C>, and Scalar<C>.  But I guess I shouldn't assume that :)

compiler/analysis-passes/check-types-Lnodca.ss
kmillikin: OK.  I guess here it's `supports-arithmetic?`.

compiler/analysis-passes/infer-types.ss
kmillikin: The code is defensive at this point.  The curve field and point types are injected into the standard library for ZKIR v3 in `zkir-v3-natives.ss` and not even present in the standard library for ZKIR v2.  So there should be no way for Compact code to construct any of these types in the v2 backend.

compiler/analysis-passes/infer-types.ss
kmillikin: I went with `valid-length?`.

compiler/analysis-passes/infer-types.ss
kmillikin: I think it's only here that we actually prevent it from happening.  The user could write a cast in either direction.

compiler/circuit-passes/check-types-Linlined.ss
kmillikin: Agreed, it would save a lot of boilerplate.

compiler/circuit-passes/flatten-datatypes.ss
kmillikin: Agreed.

compiler/circuit-passes/optimize-circuit.ss
kmillikin: That would be an improvement.

compiler/typescript-passes/print-typescript.ss
kmillikin: Yeah, and it has the advantage of keeping the naming systematic.  Another advantage that there are fewer places in the code to confuse `k1` and `r1`.

It then dispatches on the curve type twice if we continue to handle `curve-jubjub` as an exception to the naming rule.

kmillikin: I can make it fully systematic, as long as we can tolerate possibly generating undefined `CompactTypeJubjubBase` (which should be impossible) and we define `CompactTypeJubjubScalar` (for now it could simply be an alias for `CompactTypeField` to avoid introducing a change there).

compiler/typescript-passes/print-typescript.ss
kmillikin: Yes, done.

compiler/typescript-passes/print-typescript.ss
kmillikin: Yes, done.

compiler/typescript-passes/print-typescript.ss
kmillikin: `not-implemented`.  Leaving it in reminds us (me) to add test coverage of that line when the implementation is added.

compiler/typescript-passes/print-typescript.ss
kmillikin: Yeah.  I'll do that when I add the Curve25519 operations.

compiler/field.ss
kmillikin: I seem to remember either a check or a test, but I can't find it right now.  I'll ask about the details.

compiler/zkir-v3-natives.ss
kmillikin: I'd rather not.  That way we stick to the discipline of only adding shippable features to our dev branch `main`.

compiler/zkir-v3-natives.ss
kmillikin: Yeah, I agree.  The JS runtime name is more like an implementation so it should follow the signature.

runtime/src/built-ins.ts
kmillikin: Yes and the JS code will be just about the same.  Our handling of natives is too simple to support that right now, the JS implementation is just a function name with the corresponding signature, and not a code template.

runtime/src/constants.ts
kmillikin: Yeah, it would be good to have a single specification for all of these.

PR #752, opened 2026-08-31
fix(compact-update): check for compactc binary, not just directory

### LINKED ISSUE #739: compact update treats a version as installed after a failed extraction, silently skipping re-extraction on retry and masking the real error
### Context & versions:

Found during internal Hackathon:

- `compact` (toolchain-manager CLI, `tools/compact/`) v0.5.2, installed via
  `compact-installer.sh`
- Target: `compact update 0.31.1` for `x86_64-unknown-linux-musl`
- OS: Ubuntu 24.04 (noble) under WSL2 on Windows 11 — a stock image, which does not ship
  `unzip` by default
- Discovered while following `shieldedtech/shielded-dapp-template`'s documented setup flow
  (`docs/UPSTREAM.md` / `npm start`), which shells out to `compact update <pinned-version>`

### Steps to reproduce:

1. On a machine/distro that has `curl`, `tar`, `xz`, `zstd`, `gzip` but **not** `unzip`
   (e.g. a fresh Ubuntu 24.04 WSL install).
2. Install the toolchain manager via the published installer script.
3. Run `compact update 0.31.1` (or any version whose release artifact for this platform is a
   `.zip`). It fails during extraction.
4. `sudo apt install -y unzip` to install the actual missing dependency.
5. Run `compact update 0.31.1` again, **without** first removing
   `~/.compact/versions/0.31.1`.
6. Observe it fails again, differently, even though the real blocker (`unzip`) is now fixed.
7. Only `rm -rf ~/.compact/versions/0.31.1 && compact update 0.31.1` succeeds.

### Actual behavior:

- Step 3 fails with: `Failed to spawn artifact extraction command` /
  `No such file or directory (os error 2)`. The message never names which command it tried to
  spawn (in this case, `unzip`), so there's no actionable signal pointing at the missing
  dependency.
- That failed attempt leaves `~/.compact/versions/0.31.1/x86_64-unknown-linux-musl/` on disk
  containing only the raw, un-extracted `artifact.zip` — no `compactc` binary.
- Step 5 then prints `compact: x86_64-unknown-linux-musl -- 0.31.1 -- already installed` —
  i.e. it treats the *existence* of that directory as a completed install, and never retries
  extraction even though `unzip` is now available and would succeed.
- It then immediately fails on a downstream step with
  `Error: Failed to update / Caused by: Expecting a file: ".../compactc"` — a message that
  reads like a packaging/release problem, not like "your own local state is stale," so it
  actively misdirects troubleshooting.

### Expected behavior:

Either:
- the "already installed" check should verify the actual expected binary
  (`.../compactc`) exists, not just that the version directory exists, and transparently
  repair (re-extract or re-download) when it's missing; or at minimum
- the initial extraction failure should name the command it tried to spawn, so a missing
  system dependency is diagnosable without trial and error.

Ansh-Sonkusare:
## Problem

`compact update <version>` performs a quick local check before hitting the network:

```rust
if dir.exists() {
    // treat as already installed, return early
}
```

If a previous install was **interrupted** after the version directory was created but before extraction finished, `dir.exists()` returns `true` — but the `compactc` binary is missing. The tool then silently reports "already installed" and exits without downloading, leaving the toolchain in a broken state.

Issue #739 walks through the concrete reproducer: a fresh Ubuntu 24.04 WSL install lacks `unzip`, `compact update 0.31.1` fails during extraction leaving a directory containing only `artifact.zip`, and the next retry (even after `apt install unzip`) short-circuits with "already installed" and then dies downstream with the misleading `Expecting a file: ".../compactc"` error.

Fixes #739.

## Fix

Check for the binary itself instead of the directory:

```rust
if dir.join("compactc").is_file() {
```

This is the condition that actually matters: the binary is present and usable. A directory without the binary triggers the normal download-and-extract path.

This also makes the early-out consistent with the main install path below (`compact.rs:223`), which already gates on `compiler.path_compactc().is_file()` — same predicate, resolved to the same absolute path via `Compiler::create`.

## Test

Adds `test_compact_update_partial_install_not_treated_as_installed` in `tools/compact/tests/test_update.rs`. It:

1. Creates a temp `--directory` with `versions/0.19.0/<target>/artifact.zip` pre-populated (no `compactc`) — mirroring the partial-install state
2. Runs `compact update 0.19.0`
3. Asserts the output matches the existing `err_old_version.txt` golden file (i.e. control reached `resolve_version`, which then rejects `0.19.0` as too old)

Verified locally:

- [x] Test **passes** with the fix applied
- [x] Test **fails** with the fix reverted (`dir.exists()` restored) — output becomes `compact: <target> -- 0.19.0 -- already installed` followed by the exact misleading `Expecting a file: ".../compactc"` error the issue calls out, confirming both the bug and the reproducer
- [x] Uses the pre-existing `err_old_version.txt` fixture — no new golden file needed
- [x] Does not require `unzip` (invalid version bails before extraction), so unlike some existing update tests it's environment-independent

### DIFF
--- a/tools/compact/src/bin/compact.rs
+++ b/tools/compact/src/bin/compact.rs
@@ -177,7 +177,7 @@ async fn update(cfg: &CommandLineArguments, command: &UpdateCommand) -> Result<(
             .join(version.to_string())
             .join(cfg.target.to_string());
 
-        if dir.exists() {
+        if dir.join("compactc").is_file() {
             let compiler = Compiler::create(cfg, version.clone(), cfg.target).await?;
 
             println!(
--- a/tools/compact/tests/test_update.rs
+++ b/tools/compact/tests/test_update.rs
@@ -18,6 +18,7 @@ use crate::common::{
 };
 use std::collections::HashMap;
 use std::env;
+use std::fs;
 
 mod common;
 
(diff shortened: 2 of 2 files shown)


Ansh-Sonkusare:
@JosephDenman Can you review this when you can.

PR #754, opened 2026-09-01
Fix `compact update` leaving a broken installation that reports success

### LINKED ISSUE #739: compact update treats a version as installed after a failed extraction, silently skipping re-extraction on retry and masking the real error

JosephDenman:
Fixes #739.
 
## The bug
 
On a machine without `unzip`, `compact update 0.31.1` fails to extract. Installing `unzip` and running it again does not help: the retry reports `already installed`, skips extraction, and dies with `Expecting a file: '.../compactc'`. The only way out is `rm -rf ~/.compact/versions/0.31.1`.
 
The exact-version fast path tested whether the *version directory* existed. A failed extraction leaves that directory holding nothing but `artifact.zip`, so the retry saw a directory, called it installed, and never looked for a compiler.
 
It is worse than reported. `set_current_compiler` wrote the default-compiler symlink **before** validating it, and `fs::symlink` will happily point at a path that does not exist. So the failed retry left a dangling `~/.compact/bin/compactc`, and from then on `compact check`, `compact list --installed` and `compact compile` all failed the same way. One incomplete install broke the whole installation.
 
## What changed
 
**The reported defect**
 
- The fast path requires `compactc`, not just its directory, and says so when it is repairing a previous attempt.
- `set_current_compiler` refuses to link a compiler binary that is not there, so a failure can no longer poison the other commands.
- A default compiler that is already dangling now reports the remedy — `Run 'compact update 0.31.1' to reinstall it` — rather than a bare `Expecting a file`. Upgrading alone does not repair an installation that is already broken, so the people who hit this need to be told what to run.

**Presence is not validity**
 
Requiring `compactc` rather than the directory moves the check one level down; it does not make it a completeness check. An extraction that failed *after* writing that one file still left something that looked installed:
 
```
$ compact update 0.31.1        # 27-byte compactc, extraction never finished
compact: x86_64-unknown-linux-musl -- 0.31.1 -- already installed
compact: x86_64-unknown-linux-musl -- 0.31.1 -- default.
$ compact compile --version
Error: Failed to run compactc
```
 
Archives are now unpacked into a staging directory inside the version directory and moved into place with the compiler binary **last**. `rename` is atomic within a filesystem and staging is a child of the destination, so `compactc` appears only once everything else is already there, and a failure anywhere leaves the directory without it — which the fast path treats as "not installed" and retries.
 
**No more external `unzip`**
 
The reporter's machine had `curl`, `tar`, `xz`, `zstd` and `gzip` but not `unzip`, which is a separate package that nothing else pulls in. Naming the missing command made the failure legible; it did not make the machine work. The archive is now read in-process with the `zip` crate.
 
- `default-features = false, features = ["deflate-flate2"]` — the launcher never writes an archive, so it needs neither a compressor nor a second inflate backend.
- Three new packages (`zip`, `arbitrary`, `derive_arbitrary`); `flate2` and `miniz_oxide` were already present via reqwest. Nothing binds to C, which matters for the musl cross-builds.
- The release archive is built by `zip --junk-paths`, so it is a flat set of regular files. Entries are written by file name alone, which also means a hand-made archive cannot direct a write outside the destination.
- `unzip` restored each entry's recorded mode and nothing does that for us now, so the mode is applied explicitly — a `compactc` without its executable bit is an installation that looks complete and cannot run.
**Corrupt archives repair themselves**
 
The archive is downloaded only when it is not already on disk, so a complete-but-unreadable one failed every retry identically. Failures are now classified: an unreadable zip, an undecompressable entry, or a write failing with `InvalidData` (how a checksum mismatch surfaces) is the archive's fault, and the archive is discarded and fetched once more. A full disk or a permission error propagates untouched — fetching the file again would not fix it, and discarding a good archive would throw away the whole download.
 
## Test-suite repairs, and one warning
 
**Before this branch, `cargo test` deleted your installed compilers.** `test_clean.rs` ran `clean` and `clean --keep-current` with no `--directory`, so they operated on the real `~/.compact`. They are named `*_nothing_installed` because whoever wrote them had nothing installed. Running the suite removed 0.27.0, 0.29.0 and 0.30.0 from a maintainer's machine while preparing this PR. Eight tests in total read or wrote the real installation; all are now isolated.
 
Four more expectations passed in CI and failed for real developers — CI green was actively misleading about whether this suite works:
 
| expectation | depended on |
|---|---|
| `test_compact_check_no_param` | whether you had compilers installed |
| help output | the length of `$HOME` and `$TMPDIR`, and terminal width |
| version lists | the published channel, last refreshed at 0.31.1 |
| `test_compact_self_check` / `_update` | whether you had ever installed `compact` |
 
Fixed by isolating the compact directory, pinning `COLUMNS` in the harness so help is a function of its content alone, refreshing the constants to 0.34.0/0.31.1, and pinning `AXOUPDATER_CONFIG_PATH` so "no install receipt" is true by construction.
 
`output/compile/compact_help.txt` is a copy of *compactc's* help text, so it drifts whenever `LATEST_COMPACTC_VERSION` moves; 0.34.0 documents `--feature-zkir-v3`.
 
## Toolchain and CI
 
- `rust-toolchain.toml` pins the MSRV already declared in `Cargo.toml`. `cargo clippy -- -D warnings` passed or failed depending on which clippy happened to be installed — 1.88 reported three `uninlined_format_args` failures that 1.95 did not — because `compact-test.yml` asks for `stable`, making the result a function of the calendar. Cargo's feature resolution is version-dependent too, which is how `clap`'s `wrap_help` came to be enabled by accident (below).
- `clap`'s `wrap_help` is now requested explicitly. It had been arriving only through feature unification with the `cargo-nextest` dev-dependency, so wrapped help appeared under `cargo test` while the installed binary printed one long line — the expected-output files recorded output no user had ever seen. **This changes `compact --help` for users**: it now wraps to the terminal.
- Dropped `cargo-nextest` from dev-dependencies. A dev-dependency is a library to link, not a tool to run, so it never put `cargo nextest` on anyone's PATH; CI installs the binary through `taiki-e/install-action`. It did pull nextest's tree into every test and `--all-targets` build. **The lockfile goes from 490 packages to 310.**
- Dropped `pretty_assertions`, which was listed in both `[dependencies]` and `[dev-dependencies]` and imported nowhere; the first copy linked a test-output crate into the shipped binary.
- The cargo caches hashed `tools/compact/Cargo.lock` and cached `tools/compact/target`. Neither path exists — `tools/compact` is a workspace member — so `hashFiles` matched nothing, the keys were constant and never invalidated when dependencies changed, and the target cache archived an empty directory.
## Testing
 
`cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and the full suite pass, including the network-dependent tests against the live GitHub API.
 
New coverage is 15 unit tests on extraction and installation plus 3 integration tests. The ones carrying real weight were checked by breaking the code and confirming the right test failed:
 
- removing the mode restoration → `the_extracted_compiler_is_executable`
- removing the staging step → `a_failed_installation_leaves_no_compiler_in_place`
- accepting an entry name that escapes the archive → `an_entry_whose_name_escapes_the_archive_is_refused`
- removing the deflate feature → `extracts_a_deflated_archive`
- blaming every failure on the archive, or none → the classification tests, both directions
The deflate one matters because every other extraction test stores its entries uncompressed; without a genuinely deflated archive the suite would stay green while every real installation failed.
 
## Notes for review
 
- **Needs the `skip-changelog` label.** The entry is in `tools/compact/CHANGELOG.md`, which is where every CLI entry lives; root `CHANGELOG.md` has never carried one. `changelog-check.yml` greps for root `CHANGELOG.md` plus a `compiler/compiler-version.ss` bump, so any CLI-only PR must either misfile its entry or bump the compiler for a change that does not touch it. A one-line workflow change accepting `tools/compact/CHANGELOG.md` would close that; it is deliberately not in this PR.
- The large `Cargo.lock` diff is `cargo-nextest` coming out, not a dependency change. Aside from the three `zip` additions, no dependency was added or upgraded.
- Reinstalling a version now **replaces** entries rather than merging the archive over whatever was there, since `rename` will not replace a non-empty directory. This is a change from `unzip -o`'s behaviour and is what makes reinstalling over a read-only `compactc` work — nix store files are mode 555, and replacing by rename needs no write permission on the file.
- The pin lands everyone on 1.88.0. If you would rather raise the MSRV, `channel` and `rust-version` move together.

### DIFF
--- a/tools/compact/CHANGELOG.md
+++ b/tools/compact/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The toolchain archive is now unpacked by `compact` itself rather than by
+  running `unzip`. `unzip` is not installed by default on many systems -- the
+  machine in issue #739 had `curl`, `tar`, `xz`, `zstd` and `gzip` but not
+  `unzip`, so the toolchain could not be installed at all until a package was
+  added. There is no longer an external program to install.
+
 - `--help` output now wraps to the width of the terminal in the installed
   binary, not only under `cargo test`. `clap`'s `wrap_help` feature was
   never requested by this crate: it arrived through Cargo feature
--- a/tools/compact/CHANGELOG.md
+++ b/tools/compact/CHANGELOG.md
@@ -15,6 +15,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `unzip`, so the toolchain could not be installed at all until a package was
   added. There is no longer an external program to install.
 
+- An archive that cannot be unpacked is now discarded and fetched again. The
+  archive was only downloaded when it was not already on disk, so a complete
+  but unreadable one -- a bad checksum, a resume stitched across a changed
+  artifact -- failed every retry identically, with deleting it by hand the only
+  way forward. Only the archive's own faults are retried: a full disk or a
+  permission error is reported as it happens, since fetching the file again
+  would not help.
+
 - `--help` output now wraps to the width of the terminal in the installed
   binary, not only under `cargo test`. `clap`'s `wrap_help` feature was
   never requested by this crate: it arrived through Cargo feature
--- a/tools/compact/CHANGELOG.md
+++ b/tools/compact/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `--help` output now wraps to the width of the terminal in the installed
+  binary, not only under `cargo test`. `clap`'s `wrap_help` feature was
+  never requested by this crate: it arrived through Cargo feature
+  unification with the `cargo-nextest` dev-dependency, so wrapped help
+  appeared in test builds while the binary users install printed one long
+  line. The expected-output files recorded the wrapped form, so the help
+  tests asserted output no user had ever seen. The feature is now
+  requested explicitly.
+
 ### Fixed
 
 - `compact update <version>` no longer treats a version as installed when only
--- a/tools/compact/CHANGELOG.md
+++ b/tools/compact/CHANGELOG.md
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `compact update <version>` no longer treats a version as installed when only
+  the version directory is present. An extraction that fails leaves that
+  directory behind holding nothing but `artifact.zip`, and the previous check
+  looked no further: it reported "already installed", skipped re-extraction, and
+  then failed with `Expecting a file: '.../compactc'`. Installing the missing
+  dependency and re-running now repairs the installation rather than reporting
+  the same error again.
+
+  Because the default-compiler symlink was written before it was validated, a
+  failed retry also left a dangling `~/.compact/bin/compactc`, which made every
+  later `compact check`, `compact list --installed` and `compact compile` fail
+  the same way. `compact` now refuses to link a compiler binary that is not
+  there, so one incomplete install can no longer break the others.
+
+  Fixes issue #739.
+
+- The error raised when the archive extraction command is missing now names it.
+  A machine without `unzip` previously reported only `Failed to spawn artifact
+  extraction command` followed by `No such file or directory (os error 2)`,
+  identifying neither the command that was missing nor what to do about it.
+
+- Archive extraction now passes `-o`. Standard input is closed for the
+  extraction command, so an archive left partially extracted by an interrupted
+  run produced an overwrite prompt that failed the retry instead of repairing
+  it.
+
 ## [Compact tools 0.5.2]
 
 ### Fixed
--- a/.github/workflows/compact-test.yml
+++ b/.github/workflows/compact-test.yml
@@ -45,7 +45,7 @@ jobs:
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('tools/compact/Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
           restore-keys: |
             ${{ runner.os }}-cargo-
 
@@ -90,15 +90,15 @@ jobs:
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('tools/compact/Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
           restore-keys: |
             ${{ runner.os }}-cargo-
 
       - name: Cache target directory
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
-          path: tools/compact/target
-          key: ${{ runner.os }}-target-${{ hashFiles('tools/compact/Cargo.lock') }}
+          path: target
+          key: ${{ runner.os }}-target-${{ hashFiles('Cargo.lock') }}
           restore-keys: |
             ${{ runner.os }}-target-
 
--- a/.github/workflows/compact-test.yml
+++ b/.github/workflows/compact-test.yml
@@ -55,6 +55,37 @@ jobs:
       - name: Run Clippy
         run: cargo clippy --all-targets --all-features -- -D warnings
 
+  # `rust-toolchain.toml' pins the MSRV, and cargo honours it whatever the
+  # runner installed -- so the checks above are pinned too, and lints added by
+  # later stable releases would go unseen until the pin moves. This job is the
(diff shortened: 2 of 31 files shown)


tools/compact/CHANGELOG.md
```diff
@@ -7,6 +7,59 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+  Because the default-compiler symlink was written before it was validated, a
+  failed retry also left a dangling `~/.compact/bin/compactc`, which made every
+  later `compact check`, `compact list --installed` and `compact compile` fail
+  the same way. `compact` now refuses to link a compiler binary that is not
+  there, so one incomplete install can no longer break the others.
+
+  Fixes issue #739.
+
+- The error raised when the archive extraction command is missing now names it.
```
pataei, line 53: This sentence reads weird. Please rewrite.

pataei (commented):
Generally, LGTM. We should somehow track when we need to cut a release of the dev tool. Given that it's not time consuming I can see myself advocating for a release after a fix (e.g., once this goes in, cut a release of 0.5.3). Alternatively, we can consider as part of the process of releasing the toolchain to see if the dev tool has an unreleased feature. I like the former as it releases bug fixes like this faster, however, it is more burdensome. 

The followings are mentioned by AI:
**The MSRV pin knowingly regresses a CI capability, with no follow-up filed.**
As `0768f49` mentions:
```
Raise channel' and rust-version' together. Note that this makes the Install Rust toolchain' step in compact-test.yml' advisory: cargo honours rust-toolchain.toml' regardless, so new lints from later stable releases will no longer be seen until the pin moves. A separate job on stable', allowed to fail, is the usual way to keep that signal.
```
You should either add an issue to track this or do the "add a separate job on stable, that is allowed to fail.

**The extraction path is exactly the place I want more paranoia than the PR shows.**
This code downloads an archive from a network source and unpacks a binary that will later be executed. The PR:

- Adds path-traversal protection (an_entry_whose_name_escapes_the_archive_is_refused) — good, but "escapes the archive" via ../ is only one flavor of zip-slip. What about entries that are symlinks pointing outside the destination? zip --junk-paths producing "a flat set of regular files" is true for their own release archives, but the corrupt-archive-retry logic implies the code must tolerate reading archives that aren't necessarily well-formed. Is symlink-entry handling actually excluded, or just not mentioned?
- No zip-bomb / decompression-ratio guard is mentioned anywhere. A compact update that fetches from an attacker-controlled or compromised release asset (supply-chain compromise of the GitHub release, not a hypothetical category) can currently write a small crafted archive that decompresses to something enormous, and nothing in this description suggests output size is bounded.
- More importantly: is the downloaded archive checksummed/verified against anything published, or just handed straight to the zip reader? The "corrupt archive" detection only classifies read/decompress failures — it says nothing about whether a syntactically valid but tampered archive would be caught. For a tool that installs and runs a compiler binary, "we can tell if the zip is malformed" is a much weaker property than "we verified this is the artifact GitHub published."

My take: this is a fair statement and we should consider it but not necessarily in this PR. Perhaps open a new ticket for it.

pataei (approved):
Whoops, I meant to give it an approval in my previous review.

Issue #703, opened 2026-08-07
[compact-launcher@0.5.1] version channel does not list 0.33.0 rc releases — compact update 0.33.0-rc.2 has nothing to fetch


tomtblock:
- **Environment:**
  - Launcher: `compact` 0.5.1 (installed via `compact-installer.sh` from midnightntwrk/compact `compact-v0.5.1`), Ubuntu Linux x86_64 (musl target), Node 22.23.2, Docker 29.7.2
  - Target compiler: `compactc` **0.33.0-rc.2** (`--feature-zkir-v3`), required by the sig-net/midnight-examples + `@sig-net/midnight` 0.19.0 / ledger-9 matched set
  - Matched-set context: `@midnight-ntwrk/compact-runtime` 0.18.0-rc.1, `@midnightntwrk/onchain-runtime-v4` 4.0.0-rc.3, proof server 9.0.0-rc.5_experimental
- **Phase/step:** Toolchain baseline setup for an erc20-vault-based integration build (PRD Phase 0, "install compact 0.33.0-rc.2 and pin it").
- **Expected vs actual:**
  - Expected: `compact list` shows `0.33.0-rc.2` (it exists as a GitHub release: [`compactc-v0.33.0-rc.2`](https://github.com/LFDT-Minokawa/compact/releases/tag/compactc-v0.33.0-rc.2)), and `compact update 0.33.0-rc.2` downloads and adopts it.
  - Actual: the launcher's version channel tops out at `0.31.1`:
    ```
    $ compact list
    compact: available versions

      0.31.1 - x86_macos, aarch64_macos, x86_linux, aarch64_linux
      0.31.0 - ...
      0.30.0 - ...
    ```
    None of the `0.33.0` rc releases (rc.0/rc.1/rc.2) appear, so a plain `compact update 0.33.0-rc.2` on a clean machine has nothing to fetch from the channel.
- **Minimal reproduction (deterministic):** on a machine with no `~/.compact`:
  ```bash
  curl -fsSL https://github.com/midnightntwrk/compact/releases/download/compact-v0.5.1/compact-installer.sh | sh
  compact list          # no 0.33.0* entries
  ```
- **Workaround:** pre-seed the launcher's versions dir from the GitHub release and let the launcher adopt it:
  ```bash
  RC_DIR="$HOME/.compact/versions/0.33.0-rc.2/x86_64-unknown-linux-musl"
  mkdir -p "$RC_DIR"
[...]

PR #756, opened 2026-09-01
Report the release a compiler was built from

### LINKED ISSUE #705: [compactc@0.33.0-rc.x] compact compile --version drops the prerelease suffix — rc pins are unverifiable from the printed version
**Environment:** compact launcher 0.5.1, compactc 0.33.0-rc.2 (x86_64-unknown-linux-musl, direct-download per the workaround in #703), Ubuntu Linux, Node 22. Hit while pinning a matched-set toolchain for a Stagenet+Sepolia bridge build (`tomtblock/BAHack`).

**Phase/step:** CI toolchain assertion + zk-key cache keying.

## What we hit

With `0.33.0-rc.2` installed and active, `compact compile --version` prints `0.33.0` — the same string rc.0 (or any other 0.33.0 prerelease) prints. A project pinning a specific rc cannot assert its pin from the compiler's own version output: a CI cache key or a version gate built on the printed string silently accepts the wrong rc, and an rc-to-rc regression then surfaces as an unrelated runtime failure far from the cause.

**Expected vs actual:** expected `compact compile --version` → `0.33.0-rc.2`; actual → `0.33.0`. Deterministic.

Related: the launcher's channel does not list the 0.33.0 rc's at all (#703), so the rc is direct-downloaded and adopted — making the printed version the *only* self-report, and it is lossy. (Class: silent-failure — nothing errors; the wrong compiler just gets trusted.)

## Minimal reproduction

```bash
compact update 0.33.0-rc.2   # or the direct-download + symlink adoption from #703
compact compile --version    # prints "0.33.0" — indistinguishable from rc.0
```

## Suggestion

Print the full version including the prerelease suffix (and ideally the build/commit), e.g. `0.33.0-rc.2`. `--version` output is the natural pin anchor for CI.

## Our workaround

Prove the active rc via the launcher's symlink instead: `readlink ~/.compact/bin/compactc` must resolve under `versions/0.33.0-rc.2/` — encoded in our version manifest and asserted in CI, and used as the CI zk-key cache key. (Note the launcher does not put `compactc` itself on PATH — dispatch is via `compact compile` — so a `command -v compactc`-based readlink check comes back empty; the check must go through `~/.compact/bin` directly.)

**Impact:** degraded (CI can be made safe via the symlink), but silent-failure class: any pipeline trusting the printed version accepts a mismatched rc without an error.

JosephDenman:
`compactc --version` reports the release a compiler was built from — the full tag with its prerelease identifier, plus the commit and its date — instead of the bare triple committed in the source. Closes #705.

**The mechanism.** The tag arrives as a workflow input and cannot be committed, so `release-build.yml` stamps it into `compiler/compiler-version.ss` at build time (`scripts/stamp-compiler-version.sh`) and then refuses to upload a zip whose binary does not report the stamped line. Unstamped builds report `-dev`, so nothing unstamped can pass for a release.

**The process change.** The public release copies bytes and never compiles, so the final must be built as the final: after QA approves a candidate, one more internal run at `vX.Y.Z` on the same commit, and that build is what `public-release.yml` promotes. That workflow now refuses a prerelease version; it resolves the branch it takes notes and docs from to a single commit, and separately resolves the internal run's `vX.Y.Z` tag to the commit that was built — a missing tag is a hard error, so a version the internal workflow never built cannot be promoted at all. `compactc-vX.Y.Z` lands on that released commit rather than on the branch tip, so the two tags always name the same commit, and the archive being republished is downloaded and checked to report this version and that commit before anything ships — an internal re-run that swapped bytes after QA signed off now fails loudly. The internal run likewise resolves its branch once, pushes its tag idempotently, and keys concurrency by version.

**The language version is derived rather than typed.** It reaches the release titles and the published version table in `midnight-docs`, and it was a required hand-typed input on both `public-release.yml` and `copy-release-notes.yml` with nothing to check it against, so a typo mislabelled the published docs and outlived the release. Both workflows now read `compactc --language-version` off the released binary — `public-release.yml` from the archive it already downloads for the version check — and the inputs are gone.

**Compact Tool Release no longer fires on toolchain tags.** dist generates that workflow's tag filter from `tag-namespace = "compact"`, so the filter also matched the `compactc-v*` and `compact-vsc-plugin-*` tags and started a run on every toolchain and plugin release that could only fail, because dist reads the text after the namespace as a version. Two exclusions were added to the filter.

**What readers of the output see.**

- `0.34.1-rc.2 (a1b2c3d4e 2026-09-10)`: the commit sits beside the version, not inside it as build metadata, because the version string is matched on and semver comparators do not reliably ignore metadata.
- `--version --verbose` prints release, commit, date, language and runtime versions as separate fields, `unknown` where the build recorded nothing.
- `contract-info.json` and `contract-manifest.json` gain a `compiler-commit` field (all forty characters). `compiler-version` stays valid semver and, for a final release, is byte-identical to before; strict parsers of these files will see the new field.
- The prerelease tag is printed but never compared, so a candidate satisfies the same `pragma compiler_version` constraints as its release.
- Exit flags act once the whole command line is parsed: the first one wins, and `--ledger-version --feature-zkir-v3` now honours the feature flag in either order. `format-compact --version` and `fixup-compact --version` report the commit the same way.

**Known downstream impact.** `midnight-sdk`'s manifest parser (`compact-js/compact-js/src/effect/ZKManifest.ts`) is fail-closed: any unrecognised top-level string in `contract-manifest.json` is an error, so it rejects the new `compiler-commit` key — for stamped and unstamped builds alike. That key needs registering there before a post-#756 build reaches it. `midnight-js` skips keys it does not recognise and needs no change.

**Also here.** `RELEASING.md` at the repository root — the release manual: preflight, procedure, field-by-field workflow reference, message-to-cause-to-action troubleshooting — linked from `CONTRIBUTING.md`, with matching corrections to `doc/dev-builds.md`.

**Out of scope, tracked as follow-ups.** Publishing candidates where `compact update` can see them (#703), and triggering the release from the pushed tag rather than a dispatch input.


rust-toolchain.toml
```diff
@@ -0,0 +1,30 @@
+# This file is part of Compact.
+# Copyright (C) 2026 Midnight Foundation
```
kmillikin, line 2: Not Midnight Foundation.

I don't think we have a standard notice, but this should probably be "Copyright 2025 the Minokawa project contributors" or something.

JosephDenman: "Midnight Foundation" is in almost every other license notice in this project. You're saying lets just start using "Copyright 2025 the Minokawa project contributors" from now on?

.github/workflows/compute-release-notes.yml
```diff
@@ -13,7 +13,7 @@ on:
         required: true
         type: string
       language_version:
-        description: 'Language version (e.g. v1.2.3)'
+        description: 'Language version as x.y.z -- callers read it from the released compactc, never from input'
```
pataei, line 16: Why not drop the `language_version` input all together if the caller doesn't use it?

.github/workflows/internal-release.yml
```diff
@@ -132,8 +154,12 @@ jobs:
         env:
           INTERNAL: ${{ inputs.version }}
         run: |
-          if [[ ! "$INTERNAL" =~ ^v[0-9]+\.[0-9]+\.[0-9]+-.+$ ]]; then
-            echo "::error::version must match v*.*.*-* format (e.g. v1.2.3-rc.1), got: $INTERNAL"
+          # A final release is built here too and promoted by public-release.yml, so
+          # what ships is built as the release rather than renamed. See issue #705.
```
pataei, line 158:
```suggestion
          # what ships is built as the release rather than renamed.
```

.github/workflows/internal-release.yml
```diff
@@ -334,14 +360,26 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          ref: ${{ needs.setup.outputs.branch }}
+          ref: ${{ needs.setup.outputs.sha }}
 
+      # The tag survives a failure below this point, so a re-run must tolerate its
```
pataei, line 365: Not sure I understand what this comment means. Can you pls rephrase?

.github/workflows/public-release.yml
```diff
@@ -4,15 +4,11 @@ on:
   workflow_dispatch:
     inputs:
       internal_version:
-        description: 'Internal version (e.g. v1.2.3-rc.1, v1.2.3-alpha.0)'
-        required: true
-        type: string
-      language_version:
-        description: 'Language version (e.g. v1.2.3)'
+        description: 'Internal version to promote -- a final release, no prerelease identifier (e.g. v1.2.3)'
```
pataei, line 7:
```suggestion
        description: 'Final release version to promote (e.g. v1.2.3)'
```

.github/workflows/public-release.yml
```diff
@@ -85,34 +83,51 @@ jobs:
         env:
           INTERNAL: ${{ inputs.internal_version }}
         run: |
-          if [[ ! "$INTERNAL" =~ ^v[0-9]+\.[0-9]+\.[0-9]+-(rc|alpha|beta)\.[0-9]+$ ]]; then
-            echo "::error::internal_version must match v*.*.*-(rc|alpha|beta).* format (e.g. v1.2.3-rc.1), got: $INTERNAL"
+          # This workflow copies an internal build and never compiles one, so the
+          # build being promoted has to have been made as the release -- otherwise
+          # the binary reports a candidate version under a final tag. Cut an internal
+          # release with this exact version first. See issue #705.
```
pataei, line 89:
```suggestion
          # release with this exact version first.
```

compiler/compactc.ss
```diff
@@ -36,7 +36,12 @@ circuits.
 The following flags, if present, affect the compiler's behavior as follows:
   --help prints help text and exits.
 
-  --version prints the compiler version and exits.
+  --version prints the compiler version and exits.  Builds that recorded a
+    commit follow the version with that commit and its date in parentheses.
+
+  --verbose, given with --version, prints the version, commit, commit date, and
```
pataei, line 42:
```suggestion
  --version --verbose, prints the version, commit, commit date, and
```

compiler/compactc.ss
```diff
@@ -36,7 +36,12 @@ circuits.
...
+    language and runtime versions as separate fields, showing unknown for
+    anything the build did not record.  It has no other effect.
```
pataei, line 44:
```suggestion
    anything the build did not record. 
```

compiler/compactc.ss
pataei: Can I also say `--verbose --version`? if so we can add that too, but I prefer the help being clear on the order.

compiler/compactc.ss
pataei, line 1: @dybvig pls review the `*.ss` files, mainly `compactc.ss` and `version.ss`. They LGTM but I want to make sure they come across your desk.

tests-e2e/src/resources/compiler_man_page.txt
```diff
@@ -10,7 +10,12 @@ circuits.
 The following flags, if present, affect the compiler's behavior as follows:
   --help prints help text and exits.
 
-  --version prints the compiler version and exits.
+  --version prints the compiler version and exits.  Builds that recorded a
+    commit follow the version with that commit and its date in parentheses.
+
+  --verbose, given with --version, prints the version, commit, commit date, and
```
pataei, line 16: Reminder: If you adopt the change I asked for in `compactc.ss` you need to update it here too.

CHANGELOG.md
```diff
@@ -5,6 +5,58 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Toolchain 0.34.102, language 0.26.0, runtime 0.19.101]
+
+### Fixed
+
+- `compactc --version` now reports the release it was built from, including any
+  prerelease identifier and the commit. It previously reported only the
+  major.minor.bugfix triple committed in `compiler/compiler-version.ss`, so every
```
pataei, line 14:
```suggestion
  major.minor.bugfix triple, so every candidate for a release reported that release.
```

CHANGELOG.md
```diff
@@ -5,6 +5,58 @@ All notable changes to this project will be documented in this file.
...
+  candidate for a release reported that release: a compiler built from
```
pataei, line 15: Drop the rest of this paragraph. It's too verbose.

CHANGELOG.md
```diff
@@ -5,6 +5,58 @@ All notable changes to this project will be documented in this file.
...
+  Builds that are not releases report `-dev`: the scheduled build
+  and the on-demand dev publish have no release to name, and a dev publish is
+  installable, so one reporting the same shape as a finished release could pass
+  for it. `-dev` sorts below every release of the same triple, so a version
+  check that wanted a release fails instead of passing. That is also the value
+  committed in `compiler/compiler-version.ss`, so a build nothing stamped cannot
+  pass for a release either.
+
+  The commit is reported beside the version rather than inside it --
+  `0.34.102-rc.2 (a1b2c3d4e 2026-09-10)`, the commit abbreviated and dated the
```
pataei, line 31:
```suggestion
  `0.34.102-rc.2 (a1b2c3d4e 2026-09-10)` -- and is recorded in full in `contract-info.json` and
```

CHANGELOG.md
```diff
@@ -5,6 +5,58 @@ All notable changes to this project will be documented in this file.
...
+  way `rustc --version` does it -- and is recorded in full in `contract-info.json` and
```
pataei, line 32:
```suggestion
```

CHANGELOG.md
```diff
@@ -5,6 +5,58 @@ All notable changes to this project will be documented in this file.
...
+
+  `compactc --version --verbose` reports `release`, `commit-hash`,
+  `commit-date`, `language-version` and `runtime-version` as separate fields,
+  so a script need not parse one out of the other and a bug report needs one
+  command rather than three. Fields the build did not record read `unknown`
+  rather than being dropped, so the set of fields does not depend on how the
+  compiler was built.
+
+  Release candidates still satisfy the same `pragma compiler_version`
+  constraints as the release they are candidates for -- the tag is printed but
```
pataei, line 46: I find everything after `--` too much of an explanation in this paragraph.

RELEASING.md
```diff
@@ -0,0 +1,832 @@
...
+
+## 1. Before you start
+
+### What you need
+
+- Write access to `LFDT-Minokawa/compact`: releasing is dispatching its workflows, and
+  a CLI release pushes a tag to it.
+- Read access to `midnight-ntwrk/artifacts`, to see internal releases land.
+- The workflows authenticate with organisation secrets, so no personal token is needed.
+- [#topic-release-coordination](https://shielded.slack.com/archives/C085QCBL2HF) — where
```
pataei, line 35: Hmm not sure if we should put shielded related info here such as a slack channel.

RELEASING.md
```diff
@@ -0,0 +1,832 @@
...
+### What you are releasing
+
+A Compact release is four separate things that are versioned and published
+independently. Knowing which one you are dealing with prevents most of the confusion in
+this document.
+
+| Thing | Version lives in | Published to |
+| ----- | ---------------- | ------------ |
+| **compactc**, the compiler (the toolchain) | `compiler/compiler-version.ss`, `flake.nix` | GitHub releases, three repositories (below) |
+| **The Compact language** | `compiler/language-version.ss` | Nothing of its own. It appears in release titles, in the published version table, and in `pragma language_version`. |
```
pataei, line 47: Yeah but no, we're not releasing a language per se, the language is part of the compiler. So it's not an artifact that we're releasing.

RELEASING.md
```diff
@@ -0,0 +1,832 @@
...
+
+| Repository | Who can see it | What lands there |
+| ---------- | -------------- | ---------------- |
+| `midnight-ntwrk/artifacts` | Midnight, internal | Every internal release. This is the store the public release is later copied *from*. |
+| `LFDT-Minokawa/compact` | Public — this is the source repository | Candidates, published as **prereleases** — publicly visible. Public releases as full releases. Also where all the tags live. |
+| `midnightntwrk/compact` | Public | Public releases only. **This is the one the `compact` tool reads**, so a build that is not here cannot be installed with `compact update`. |
+
+Because `compact update` only ever looks at `midnightntwrk/compact`, and only public
+releases go there, **release candidates cannot be installed with the `compact` tool**
+(issue #703). Whoever is validating a candidate gets it from the internal store or from
```
pataei, line 61:
```suggestion
Whoever is validating a candidate gets it from the internal store or from
```

RELEASING.md
```diff
@@ -0,0 +1,832 @@
...
+Every toolchain release passes through some or all of three repositories.
+
+| Repository | Who can see it | What lands there |
+| ---------- | -------------- | ---------------- |
+| `midnight-ntwrk/artifacts` | Midnight, internal | Every internal release. This is the store the public release is later copied *from*. |
+| `LFDT-Minokawa/compact` | Public — this is the source repository | Candidates, published as **prereleases** — publicly visible. Public releases as full releases. Also where all the tags live. |
+| `midnightntwrk/compact` | Public | Public releases only. **This is the one the `compact` tool reads**, so a build that is not here cannot be installed with `compact update`. |
+
+Because `compact update` only ever looks at `midnightntwrk/compact`, and only public
+releases go there, **release candidates cannot be installed with the `compact` tool**
```
pataei, line 60:
```suggestion
releases go there, **release candidates cannot be installed with the `compact` tool**.
```

RELEASING.md
```diff
@@ -0,0 +1,832 @@
...
+run resolves your branch to one commit up front and then builds, tests and tags that
+exact commit, so a push to the branch mid-run cannot change what you get; the branch
+you pick still decides which commit that is.
+
+If the runtime also needs releasing, the candidates have to be interleaved in a
+particular order — see "Releasing the runtime alongside" below before you start.
+
+### Step 6 — Announce it
+
+Post in [#topic-release-coordination](https://shielded.slack.com/archives/C085QCBL2HF)
```
pataei, line 196: Again, not a good idea to link slack here. We should really move these announcements to discord. Maybe we should have such parts (posting in slack channels) in our team slack channel as a reminder.

RELEASING.md
```diff
@@ -0,0 +1,832 @@
...
+right version on the branch, run the workflow, promote, verify.
+
+Two rules before anything else:
+
+- **Never publish an internal version publicly.** The runtime follows the two-track
+  versioning in §6, so a version shaped `0.x.10y` is a development version and must not
+  reach the public registry.
+- Every release, internal or public, needs an entry in the dependency compatibility
+  matrix, recording which ledger and onchain-runtime versions this runtime was built
+  against. Ask where the current compatibility matrix lives in the
```
pataei, line 344: Again, not a good idea to link slack. This isn't a blocker, we should include it as a note in a release page in our team slack channel and edit the matrix where ever it lives. I posted the question in the release channel so we'd know the answer.

RELEASING.md
```diff
@@ -0,0 +1,832 @@
...
+  is tagged `compactc-v<version>`.
+- For a candidate, delete the prerelease from
+  [LFDT-Minokawa/compact](https://github.com/LFDT-Minokawa/compact/releases). The
+  workflow creates it as a draft and then immediately un-drafts it, so it is publicly
+  visible. A final version does not get one.
+- Delete **both** tags from
+  [LFDT-Minokawa/compact](https://github.com/LFDT-Minokawa/compact/tags): `v<version>`
+  and, for a candidate, `compactc-v<version>`.
+
+Re-running instead of undoing: re-dispatching the same version on the same commit is
```
pataei, line 571: So the tag isn't pushed again? Keep in mind our releases are immutable. So far when you'd push the same tag again to run a release workflow again, you'd get an error the tag is already used even if it's cleaned up (removed) due to the immutable setting. So I'm wondering if you're suggesting that this new mechanism gets around the immutable setting.

pataei (commented):
Generally LGTM and it's worth trying. Pls address my comments (especially remove links to our slack channel). Happy to try this out for our next release.

compiler/compactc.ss
JosephDenman: Yes, you can do both. The help message states this now.

compiler/program-common.ss
```diff
@@ -41,10 +41,50 @@
       (fprintf (current-error-port)
         "       --help displays detailed usage information\n")))
 
-  (define (print-compiler-version)
-    (fprintf (current-output-port)
-             "~a\n"
-             compiler-version-string))
+  ;; Verbose prints every field always, `unknown` where the build recorded
```
dybvig, line 44: I'd prefer all of the supporting definitions be included within the definition of `print-compiler-version` (at the the top of the second `case-lambda` clause) or packaged together in a module that exports only `print-compiler-version`.   I don't like that they are visible to code that has no use for them.  Also, I would replace the reference to `unknown-field` with `"unknown"` and remove the definition of `unknown-field`.

scripts/stamp-compiler-version.sh
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env bash
```
dybvig, line 1: I'm not crazy about modifying compiler-version.ss.  This seems inherently fragile.  Maybe this should be modifying a `version-config.ss` file that contains all and only the lines that need to be modified, and the entire file `version-config.ss` file can just be written rather than edited.

compiler/version.ss
```diff
@@ -15,18 +15,43 @@
...
-    (fields feature major minor bugfix))
+    (fields feature major minor bugfix tag)
+    (protocol
+      (lambda (new)
+        (case-lambda
+          [(feature major minor bugfix) (new feature major minor bugfix "")]
+          [(feature major minor bugfix tag) (new feature major minor bugfix tag)]))))
+
+  ;; `version` with its tag replaced. Separate from the constructor because the
+  ;; tag is not known when the version is committed, and because five regex
```
dybvig, line 40: > five regex consumers depend on the literal ...

This is a big red flag, very fragile.  We should avoid this if at all possible.

dybvig (changes requested):
I've marked some things I would like changed.  Most importantly, I don't like modifying `compiler-version.ss` and instead prefer a separate config file that contains all and only the information that is produced by the version stamper.  I didn't look for all of the regex matches for a specific `make-version` call, but I'm hoping changing the above will eliminate the need for any of them.

compiler/program-common.ss
JosephDenman: Done. Moved the three helpers to the top of the second case-lambda clause and inlined "unknown".

compiler/version.ss
JosephDenman: Four of the five were already on main. But, agreed. I added `scripts/read_version.sh` so that the read is in one place.

scripts/stamp-compiler-version.sh
JosephDenman: Agreed, did it the way you suggested. I added `compiler/version-config.ss` that holds exactly the three stamped values, and the script writes it whole. The triple stays in `compiler-version.ss`.

pataei:
@JosephDenman Forgot to mention this in my previous review, if I remember correctly, you mentioned somewhere that this change would impact the upstream components in how they pull in Compactc (either compact.js or midnight.js). If so, we should ping them so that they can prep for it in a timely manner.

dybvig (approved):

PR #760, opened 2026-09-02
refactor(fuzzer): generalize repeated shapes and unify grammar

whankinsiv:
This PR switches the fuzzer implementation to TypeScript, unifies the grammar in one file, and generalizes repeated shapes to cut duplication, with coverage checked against the old grammar. It also validates the grammar before generating contracts, so a broken table fails the run instead of quietly producing weaker tests. 

Coverage additions:
- 14 ledger methods, including balance, block-time, unshielded-coin, and minting operations. 
- 4 block-time circuits: `blockTimeLt`, `blockTimeLte`, `blockTimeGt`, and `blockTimeGte`

Follow up PRs will expand the grammar set to include the crypto, cross-contract call, and event syntax. 

Validation:
```
cd tests-e2e
NO_OF_FUZZER_TESTS=10 npm run test:fuzzer
```


tests-e2e/src/fuzzer/grammar/compact.ts
```diff
@@ -0,0 +1,1212 @@
...
+
+/* `TERMINAL_GENERATORS` is typed against this list. */
+export const TERMINALS = [
+    'random_version', 'random_string', 'random_number', 'very_small_random_number',
+    'small_random_number', 'random_table', 'random_mixed_table',
+] as const;
+
+export type Terminal = (typeof TERMINALS)[number];
+
+export const ENTRY_POINTS = {
```
pataei: Why are `assert`, `if`, and `for` in `ENTRY_POINTS`? They can only be accessed through either a circuit or a constructor.

tests-e2e/src/fuzzer/grammar/compact.ts
```diff
@@ -0,0 +1,1212 @@
...
+ * ================================================================== */
+
+const harness: Grammar = {
+    statements: [
+        ['import CompactStandardLibrary;', 'line_separator', 'statement_variables', 'statement_declaration', 'statement_body'],
+    ],
+    statement_variables: [DECLARED_ADTS.flatMap(([adt, type]) => adtDeclaration(adt, type))],
+    statement_declaration: [['constructor()'], ['export circuit test(): []']],
+    statement_body: [['{\n ', ...preambleRefs('binding_', 4), 'statement', 'after_statement', '\n}']],
+    assert_statements: [contract('constructor_harness', 'assert_body')],
```
pataei: Not sure if the purpose of this PR is the comprehensiveness of fuzzer or not, but if it is we should add a circuit harness for if, for, and assert.

pataei (commented):
LGTM, could you pls add a readme to fuzzer to lower the barrier of maintaining it?

It'd be very practical to pin the version of the language that the fuzzer is fuzzing. This way once the fuzzer is comprehensive we'd know what cases need to be added based on the changes in the language from one version to another. Not a blocker for this PR, just a suggestion moving forward.

Tested this locally with the commands mentioned in PR's body.

tests-e2e/eslint.config.mjs
```diff
@@ -32,8 +32,7 @@ export default [
     ...compat.extends('plugin:prettier/recommended', 'plugin:@typescript-eslint/recommended-requiring-type-checking'),
     {
         plugins: { vitest },
-        // fuzzer is in cjs, so ignore ts checks there for now
-        ignores: ['src/fuzzer/**/*', 'src/resources/**/*'],
+        ignores: ['src/resources/**/*'],
```
pataei: Will the fuzzer test be included in the release process? I have a vague memory of this happening and I really don't want the fuzzer finding bugs blocking our release.

tests-e2e/src/fuzzer/grammar/compact.ts
whankinsiv: Yes, a follow-up PR which @golddydev is working on will focus on bringing the fuzzer up to date with the latest features and missing syntax.

tests-e2e/eslint.config.mjs
whankinsiv: The release script skips running the fuzzer. `release-test.yml` runs `yarn test:release` which has `--exclude **/fuzzer.e2e.test.ts`.

pataei (approved):
LGTM, ty!

PR #778, opened 2026-09-18
feat: add secp256r1 recover and secp256r1 test suits

golddydev:
## Summary

This PR is to add secp256r1 recover runtime and secp256r1 related tests.

### Compiler
- Add `secp256r1` branch to `field-type-equal` in optimize circuit.

### Runtime
- Add `secp256r1EcdsaRecover` runtime util
- Fix `ToProjective` runtime utils for both curves to throw a point which is not on the curve. (the `fromAffine` only do range checks)
- Add `secp256r1.test.ts` to test secp256r1 curve fields operation, and point operation.
- Add more tests to `secp256k1.test.ts`


compiler/circuit-passes/missing-guard-workarounds.ss
```diff
@@ -69,6 +69,8 @@
                [(field-native) (> len (field-bytes))]
                [(field-base (curve-secp256k1)) #t]
                [(field-scalar (curve-secp256k1)) #t]
+               [(field-base (curve-secp256r1)) #t]
+               [(field-scalar (curve-secp256r1)) #t]
```
golddydev, line 73: Added `Secp256r1` fields, so compiler doesn't make any `if` workaround for `Secp256r1` -> Bytes<32>

compiler/circuit-passes/optimize-circuit.ss
dybvig: Please remove this definition of `field-type-equal?` and copy the definition of `same-field-type?` (along with its helper `same-curve-type?`) from `check-types/Lflattened` here and use it instead.  (The code should be shared, but we presently don't have good place to put it.)

dybvig (changes requested):

compiler/circuit-passes/missing-guard-workarounds.ss
kmillikin: We should probably use `strict-nanopass-case` to handle all field types like this:
```scheme
(strict-nanopass-case (Lflattened Field-Type) ftype
  [(field-native) (> len (field-bytes))]
  [(field-base ,ctype)
    (strict-nanopass-case (Lflattened Curve-Type) ctype
      [(curve-curve25519) #t]
      [(curve-jubjub) (assert cannot-happen)]
      [(curve-secp256k1) #t]
      [(curve-secp256r1) #t])]
  [(field-scalar ,ctype)
    (strict-nanopass-case (Lflattened Curve-Type) ctype
      [(curve-curve25519) #t]
      [(curve-jubjub) #f]
      [(curve-secp256k1) #t]
      [(curve-secp256r1) #t])])
```

compiler/circuit-passes/missing-guard-workarounds.ss
```diff
@@ -98,12 +100,14 @@
              (lambda (triv2)
                (list `(= 1 ,var-name (< ,bits ,triv1 ,triv2))))))))]
     [(bytes->field ,src ,ftype ,len ,triv1 ,triv2)
-     (nanopass-case (Lflattened Field-Type) ftype
-       [(field-base (curve-secp256k1)) (list `(= 1 ,var-name ,ir))]
-       [(field-scalar (curve-secp256k1)) (list `(= 1 ,var-name ,ir))]
-       [(field-native) (guard (<= len (field-bytes))) (list `(= 1 ,var-name ,ir))]
-       [(field-native)
-        (with-output-language (Lflattened Statement)
+     (with-output-language (Lflattened Statement)
```
kmillikin: Also here, rewrite in terms of `strict-nanopass-case` to be exhaustive (without an `else`).

compiler/circuit-passes/optimize-circuit.ss
```diff
@@ -51,6 +51,14 @@
           [(field-scalar (curve-secp256k1))
            (nanopass-case (Lflattened Field-Type) ftype^
              [(field-scalar (curve-secp256k1)) #t]
+             [else #f])]
+          [(field-base (curve-secp256r1))
```
kmillikin: And the same.  Rewrite the outer `nanopass-case` to use `strict-nanopass-case` and to be exhaustive without an `else`.

compiler/zkir-v3-natives.ss
```diff
@@ -80,8 +80,8 @@
 ;; -- Curve25519Base
 (declare-native-entry circuit neg
   "__compactRuntime.curve25519BaseNeg"
-  ([s (TypeRef Curve25519Scalar) (discloses "the negation of")])
-  (TypeRef Curve25519Scalar))
+  ([s (TypeRef Curve25519Base) (discloses "the negation of")])
```
kmillikin: Good catch.

runtime/src/utils.ts
```diff
@@ -63,14 +62,23 @@ export const toHex = (s: Uint8Array): string => Buffer.from(s).toString('hex');
 
 /**
  * Lift the simple affine `Secp256k1Point` representation into a noble-curves
- * projective point. Identity maps to `Point.ZERO`; every other input is validated
- * to lie on the curve by `fromAffine`.
+ * projective point. Identity maps to `Point.ZERO`; every other input is checked
+ * to have in-range coordinates and to lie on the curve.
+ *
+ * `fromAffine` only range-checks the coordinates, so the curve equation needs the
```
kmillikin: No, this (about `fromAffine`) is not part of the doc comment.  It's describing the implementation, not the API.

Put this comment in the implementation.

runtime/src/utils.ts
```diff
@@ -63,14 +62,23 @@ export const toHex = (s: Uint8Array): string => Buffer.from(s).toString('hex');
...
+ * separate `assertValidity` call.
  */
 export function secp256k1ToProjective(p: Secp256k1Point): ReturnType<typeof secp256k1.Point.fromAffine> {
   if (p.identity) {
     return secp256k1.Point.ZERO;
   }
-  return secp256k1.Point.fromAffine({ x: p.x, y: p.y });
+  try {
```
kmillikin: I don't know if we really want this.  It adds the overhead in JS to, for example, every curve point addition which now has two different checks even if the value was produced from an operation guaranteed to produce valid points.

We could declare this function internal and then just have it assume that it's given valid points.  The only way to construct an invalid point is to return one from a witness or pass one from JS as a circuit input, and maybe we actually want this check on those values.

compiler/circuit-passes/optimize-circuit.ss
golddydev: [b32057c](https://github.com/LFDT-Minokawa/compact/pull/778/commits/b32057c483534d6a2991c9a3cc782fdf5f51dfb6)

Fixed it

compiler/circuit-passes/missing-guard-workarounds.ss
golddydev: [4453ab7](https://github.com/LFDT-Minokawa/compact/pull/778/commits/4453ab7904b86f6ad59d25e91c7c65f0843f4795)

compiler/circuit-passes/missing-guard-workarounds.ss
golddydev: [4453ab7](https://github.com/LFDT-Minokawa/compact/pull/778/commits/4453ab7904b86f6ad59d25e91c7c65f0843f4795)

runtime/src/utils.ts
golddydev: [4194cb1](https://github.com/LFDT-Minokawa/compact/pull/778/commits/4194cb1353d347db56efb9b9c56a53933e76b3e4)

Added this for ed25519 only for now.

PR #780, opened 2026-09-21
Reject use of identity points as public keys for jubjubSchnorrVerify and secp256k1EcdsaVerify

dybvig:
See the change log entry for details.


kmillikin (approved):

kmillikin (approved):
Still looks good to me.

